PR #106, opened 2014-06-10
added promises

rstone770:
Hey. I love this awesome lib and thought i'd do my part and add something that i find incredibly useful.

I've added promises to glob which allows you follow the promise pattern when dealing with async glob searches. It just makes it pretty. 

I wasn't really sure how to write tests for it so let me know.

Thanks and keep up the great work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,23 @@ Perform an asynchronous glob search.
 
 Perform a synchronous glob search.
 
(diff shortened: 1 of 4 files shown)


cbartondock:
It's 2020 and not having promise support is a horrible look. Might want to reconsider.

Issue #123, opened 2014-10-30
nocase + drive letter on Windows = match nothing


whitetrefoil:
Env:

```
Win7 x64
node v0.10.33 x64 binary version
┬ glob@4.0.6
── graceful-fs@3.0.4
── inherits@2.0.1
─┬ minimatch@1.0.0
 ├── lru-cache@2.5.0
 └── sigmund@1.0.0
```

I have a directory `C:\Temp`.  2 files there: `t.txt` & `t2.txt`.
I have a JS file `D:\tmp\glob-test.js`:

``` javascript
var glob = require('glob');

glob('C:/Temp/*.txt', function (err, match) {
  console.log('case');
  console.log(match);
});

glob('C:/Temp/*.txt', {nocase: true}, function (err, match) {
  console.log('nocase');
  console.log(match);
});
```

I open CMD, `cd` to `D:\tmp`, run `node glob-test.js`.  It output:

```
D:\tmp>node glob-test.js
case
[ 'C:/Temp/t.txt', 'C:/Temp/t2.txt' ]
nocase
[]
```

kristinochka:
The issue is still present in the latest version of glob.

isaacs:
Drives are first class citizens now in v9.

PR #146, opened 2014-12-02
Add support for UNC paths

### LINKED ISSUE #74: Issue with UNC paths on Windows
When a UNC path is provided as part of the pattern to match, it is incorrectly transformed to an invalid path on the local computer.

e.x. `\\nas\node\` incorrectly becomes `/nas/node` which is treated as `C:\nas\node\`.

staticshock:
Wildcards that now work: \host\directory*

Wildcards that still don't work: \host*

Fixes #74 and #123

### DIFF
--- a/common.js
+++ b/common.js
@@ -15,18 +15,27 @@ var path = require("path")
 var minimatch = require("minimatch")
 var Minimatch = minimatch.Minimatch
 
-function absWin (p) {
-  if (absUnix(p)) return true
+function WinPath (p) {
+  if (!(this instanceof WinPath))
+    return new WinPath(p)
+
   // pull off the device/UNC bit from a windows path.
   // from node's lib/path.js
   var splitDeviceRe =
       /^([a-zA-Z]:|[\\\/]{2}[^\\\/]+[\\\/]+[^\\\/]+)?([\\\/])?([\s\S]*?)$/
   var result = splitDeviceRe.exec(p)
-  var device = result[1] || ''
-  var isUnc = device && device.charAt(1) !== ':'
-  var isAbsolute = !!result[2] || isUnc // UNC paths are always absolute
+  this.device = result[1] || ''
+  this.sep = result[2] || ''
+  this.tail = result[3] || ''
+  this.isUnc = !!this.device && this.device.charAt(1) !== ':'
+  this.isAbsolute = !!this.sep || this.isUnc // UNC paths are always absolute
+}
 
-  return isAbsolute
+function absWin (p) {
+  if (absUnix(p)) return true
+
+  var winPath = new WinPath(p)
+  return winPath.isAbsolute
 }
 
 function absUnix (p) {
@@ -81,6 +90,14 @@ function setopts (self, pattern, options) {
   else {
     self.cwd = options.cwd
     self.changedCwd = path.resolve(options.cwd) !== cwd
+  }
+
+  if (process.platform === "win32") {
+    var winPath = new WinPath(pattern)
+    if (winPath.isAbsolute) {
+      options.root = winPath.device
+      pattern = winPath.sep + winPath.tail
+    }
   }
 
   self.root = options.root || path.resolve(self.cwd, "/")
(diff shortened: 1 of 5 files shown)


isaacs:
Needs a test that fails without this patch, and passes with this patch.

staticshock:
Yep, here you go. I'm not really good at node.js stuff, so let me know if this is on the right track. I have no idea if it's possible to create network shares without launching a subprocess, for instance. Also, no idea if global-leakage.js is going haywire here. Right now I'm seeing this on OS X, and I don't know what it means:

```
assert.js:93
  throw new assert.AssertionError({
        ^
AssertionError: "ArrayBuffer:Buffer:DTRACE_HTTP_CLIENT_REQUEST:DTRACE_HTTP_CLIENT_RESPONSE:DTRACE_HTTP_SERVER_REQUEST:DTRACE_HTTP_SERVER_RESPONS == "ArrayBuffer:Buffer:DTRACE_HTTP_CLIENT_REQUEST:DTRACE_HTTP_CLIENT_RESPONSE:DTRACE_HTTP_SERVER_REQUEST:DTRACE_HTTP_SERVER_RESPONS
    at process.<anonymous> (/Users/abacker/Dropbox/Projects/node-glob/test/global-leakage.js:10:10)
    at process.emit (events.js:95:17)
not ok test/unc-paths.js ................................ 0/1
    Command: "/usr/local/Cellar/node/0.10.36/bin/node unc-paths.js"
    TAP version 13
    not ok 1 test/unc-paths.js
      ---
        exit:    8
        stderr:  |
          assert.js:93
            throw new assert.AssertionError({
                  ^
          AssertionError: "ArrayBuffer:Buffer:DTRACE_HTTP_CLIENT_REQUEST:DTRACE_HTTP_CLIENT_RESPONSE:DTRACE_HTTP_SERVER_REQUEST:DTRACE_HTTP_SERVER_RESPONS == "ArrayBuffer:Buffer:DTRACE_HTTP_CLIENT_REQUEST:DTRACE_HTTP_CLIENT_RESPONSE:DTRACE_HTTP_SERVER_REQUEST:DTRACE_HTTP_SERVER_RESPONS
              at process.<anonymous> (/Users/abacker/Dropbox/Projects/node-glob/test/global-leakage.js:10:10)
              at process.emit (events.js:95:17)
        command: "/usr/local/Cellar/node/0.10.36/bin/node unc-paths.js"
      ...

    1..1
    # tests 1
    # fail  1
```

staticshock:
Ah, I guess global-leakage.js was testing for the presence 'TAP_Global_Harness'. Fixed in the new version.

isaacs:
Awesome, thanks.

My concern is that it's only running the test on Windows, and since most of the development (and ci) is running on unix, this won't catch it if I later break this functionality.  But now, since the tests pass on your windows system, at least I believe that it works ;)

A better approach might be to mock out the parts of the system so that the functionality can be tested on a Unix system as well.  The general pattern would be something like:

```javascript
// pretend to be windows
process.platform = 'win32'
// load glob before we muck around with fs's guts
var glob = require('../')
// mock out whatever we need to
var originalStat = fs.stat
fs.stat = function (path, cb) {
  // do whatever we want, based on the path
}
var originalReaddir = fs.readdir
fs.readdir = function (path, cb) {
  // whatever
}

test('wildcards in unc path', function (t) {
  glob('\\\\whatever\\*', function (er, results) {
    // make sure this is what we expect
    t.end()
  })
})
```

heikki:
@isaacs There's [appveyor](http://www.appveyor.com/) if you want to run tests on windows too.

- http://www.appveyor.com/docs/lang/nodejs-iojs
- https://github.com/paulmillr/chokidar/blob/master/appveyor.yml

staticshock:
I worry that I might mock something inexactly, and the tests would pass on unix and somehow still fail on Windows anyway. For instance, how should I simulate a case-insensitive stat call on a potentially case sensitive file system?

staticshock:
@isaacs, I explored your suggestion briefly, and I don't think mocking the `fs` calls plays nicely with the current implementation. Specifically, I ran into a problem mocking `path.resolve`, and haven't figured out a solution yet. I have a separate ["work in progress" branch](https://github.com/staticshock/node-glob/commits/unc-path-tests) that demonstrates the attempt, and it doesn't really work on Windows or Linux.

isaacs:
No longer an issue in v9.

Issue #70, opened 2013-07-05
Streaming?


contra:
Just proposing the idea of making glob.Glob a stream instead of an EventEmitter that pretty much is a stream.

I've written a quick wrapper to do what I need but having somebody who knows what they are doing would probably be better.

Thoughts?

ilanbiala:
What effect would this have usage-wise?

isaacs:
Rather than land this in this module, here's a quick wrapper that anyone can use

```javascript
// copyright Isaac Z. Schlueter. Public Domain.
var glob = require('glob')
var Readable = require('stream').Readable
var util = require('util')
util.inherits(GlobStream, Readable)
function GlobStream (pattern, options) {
  Readable.call(this, options)
  var self = this
  this.glob = glob(pattern, options)
  this.glob.on('match', function (m) {
    self.push(m)
  })
  this.glob.on('end', function () {
    self.push(null)
  })
  this.glob.on('abort', function () {
    self.push(null)
  })
  this.glob.on('error', function (er) {
    self.emit('error', er)
  })
})
```

PR #151, opened 2015-01-02
New: Added ignore option (fixes #115)

### LINKED ISSUE #115: consider to support options.ignore
In my opinion, it might be so useful feature.
```js
Glob('src/**', {ignore: ['**/*.compiled.js', '**/*.settings.js']})
```
Especially in cases like gulp-tasks or any cli tools working with files.
It is not the same how to filter glob results after globbing done, because of glob is not needed to find matches in ignored subdirectories (ignore: ```**/dir/**```)

vegetableman:
fixes #115 

In addition to ```minimatch``` check, the code checks if a particular pattern ends with a glob star ```**```, if it does, adds the directory to the ignore list and stops traversal for the directory matching the pattern. 

Also, if, for example, ```a/**``` is added to the ignore list, it removes ```a``` from the resulting match list.

### DIFF
--- a/common.js
+++ b/common.js
@@ -6,6 +6,7 @@ exports.ownProp = ownProp
 exports.makeAbs = makeAbs
 exports.finish = finish
 exports.mark = mark
+exports.isIgnored = isIgnored
 
 function ownProp (obj, field) {
   return Object.prototype.hasOwnProperty.call(obj, field)
@@ -41,6 +42,28 @@ function alphasort (a, b) {
   return a.localeCompare(b)
 }
 
+function ignoreMap (self, options) {
+  self.ignore = options.ignore || []
+
+  if (!Array.isArray(self.ignore))
+    self.ignore = [self.ignore]
+
+  if (self.ignore.length) {
+    self.ignore = self.ignore.map(function (pattern) {
+      if (pattern.indexOf('!') >= 0 )
+        throw new Error("Negations are not allowed in ignore")
+
+      var endsWithGlobStar = pattern.slice(-3) === '/**'
+      var endsWithStar = pattern.slice(-2) === '/*'
+      var m = { matcher: new Minimatch(pattern) }
+
+      if (endsWithGlobStar)
+        m.gmatcher = new Minimatch(pattern.slice(0, pattern.replace(/(\/\*\*)+$/, '').length))
+
+      return m
+    })
+  }
+}
 
 function setopts (self, pattern, options) {
   if (!options)
@@ -73,6 +96,8 @@ function setopts (self, pattern, options) {
   self.cache = options.cache || Object.create(null)
   self.statCache = options.statCache || Object.create(null)
   self.symlinks = options.symlinks || Object.create(null)
+
+  ignoreMap(self, options)
 
   self.changedCwd = false
   var cwd = process.cwd()
@@ -139,6 +164,11 @@ function finish (self) {
     }
   }
 
+  if (self.ignore.length)
+    all = all.filter(function(m) {
+      return !isIgnored(self, m)
+    })
+
   self.found = all
 }
 
@@ -166,7 +196,7 @@ function mark (self, p) {
 // lotta situps...
 function makeAbs (self, f) {
   var abs = f
-  if (f.charAt(0) === "/") {
+  if (f.charAt(0) === '/') {
     abs = path.join(self.root, f)
   } else if (exports.isAbsolute(f)) {
     abs = f
@@ -174,4 +204,21 @@ function makeAbs (self, f) {
     abs = path.resolve(self.cwd, f)
   }
   return abs
+}
+
+
+// Return true, if pattern ends with globstar '**', for the accompanying parent directory.
+// Ex:- If node_modules/** is the pattern, add 'node_modules' to ignore list along with it's contents
+function isIgnored (self, path, checktrailingStar) {
+  if (!self.ignore.length)
+    return false
+
+  if (checktrailingStar)
+    return self.ignore.some(function(item) {
+      return !!(item.gmatcher && item.gmatcher.match(path))
+    })
+  else
+    return self.ignore.some(function(item) {
+      return item.matcher.match(path) || !!(item.gmatcher && item.gmatcher.match(path))
+    })
 }
--- a/glob.js
+++ b/glob.js
@@ -56,6 +56,7 @@ var setopts = common.setopts
 var ownProp = common.ownProp
 var inflight = require('inflight')
 var util = require('util')
+var isIgnored = common.isIgnored
 
 var once = require('once')
 
@@ -270,13 +271,16 @@ Glob.prototype._process = function (pattern, index, inGlobStar, cb) {
 
   var abs = this._makeAbs(read)
 
+  //if ignored, skip _processing
+  if (isIgnored(this, read, true))
+    return cb()
+
   var isGlobStar = remain[0] === minimatch.GLOBSTAR
   if (isGlobStar)
     this._processGlobStar(prefix, read, abs, remain, index, inGlobStar, cb)
   else
     this._processReaddir(prefix, read, abs, remain, index, inGlobStar, cb)
 }
-
 
 Glob.prototype._processReaddir = function (prefix, read, abs, remain, index, inGlobStar, cb) {
   var self = this
--- a/sync.js
+++ b/sync.js
@@ -14,6 +14,7 @@ var alphasorti = common.alphasorti
 var isAbsolute = common.isAbsolute
 var setopts = common.setopts
 var ownProp = common.ownProp
+var isIgnored = common.isIgnored
 
 function globSync (pattern, options) {
   if (typeof options === 'function' || arguments.length === 3)
@@ -98,12 +99,17 @@ GlobSync.prototype._process = function (pattern, index, inGlobStar) {
 
   var abs = this._makeAbs(read)
 
+  //if ignored, skip processing
+  if(isIgnored(this, read, true))
+    return
+
   var isGlobStar = remain[0] === minimatch.GLOBSTAR
   if (isGlobStar)
     this._processGlobStar(prefix, read, abs, remain, index, inGlobStar)
   else
     this._processReaddir(prefix, read, abs, remain, index, inGlobStar)
 }
+
 
 GlobSync.prototype._processReaddir = function (prefix, read, abs, remain, index, inGlobStar) {
   var entries = this._readdir(abs, inGlobStar)
--- a/README.md
+++ b/README.md
@@ -264,6 +264,7 @@ filesystem.
 * `nocomment` Suppress `comment` behavior.  (See below.)
 * `nonull` Return the pattern when no matches are found.
 * `nodir` Do not match directories, only files.
+* `ignore` Add a pattern or an array of patterns to exclude matches.
 
 ## Comparisons to other fnmatch/glob implementations
 
--- a/test/bash-results.json
+++ b/test/bash-results.json
@@ -101,6 +101,8 @@
     "./test/global-leakage.js",
     "./test/globstar-match.js",
     "./test/has-magic.js",
+    "./test/ignore-sync.js",
+    "./test/ignore.js",
     "./test/mark-sync.js",
     "./test/mark.js",
     "./test/negation-test.js",
--- a/test/ignore-sync.js
+++ b/test/ignore-sync.js
@@ -1,0 +1,151 @@
+require('./global-leakage.js')
+// Ignore option test
+// Show that glob ignores results matching pattern on ignore option
+
+var glob = require('../glob.js')
+var test = require('tap').test
+
+test('get all', function(t) {
+  var results = glob('*', {sync: true, cwd: 'a'})
+  t.same(results, ['abcdef', 'abcfed', 'b', 'bc', 'c', 'cb', 'symlink'])
+  t.end()
+});
+
+test('ignore b', function(t) {
+  var results = glob('*', {sync: true, cwd: 'a', ignore: 'b'})
+  t.same(results, ['abcdef', 'abcfed', 'bc', 'c', 'cb', 'symlink'])
+  t.end()
+});
+
+test('ignore all with first letter as b', function(t) {
+  var results = glob('*', {sync: true, cwd: 'a', ignore: 'b*'});
+  t.same(results, ['abcdef', 'abcfed', 'c', 'cb', 'symlink'])
+  t.end()
+});
+
+test('ignore b/c/d in b/c', function(t) {
+  var results = glob('b/**', {sync: true, cwd: 'a', ignore: 'b/c/d'});
+  t.same(results, ['b', 'b/c'])
+  t.end()
+});
+
+// matches based on pattern specified
+test('ignores d in b/c only if pattern starts with b/c', function(t) {
+  var results = glob('b/**', {sync: true, cwd: 'a', ignore: 'd'});
+  t.same(results, ['b', 'b/c', 'b/c/d'])
+  t.end()
+});
+
+test('ignore b/c and it\'s contents using globstar', function(t) {
+  var results = glob('b/**', {sync: true, cwd: 'a', ignore: 'b/c/**'})
+  t.same(results, ['b'])
+  t.end()
+});
+
+test('ignore, get all d but that in b', function(t) {
+  var results = glob('**/d', {sync: true, cwd: 'a', ignore: 'b/c/d'})
+  t.same(results, ['c/d'])
+  t.end()
+});
+
+test('ignore, get all a/**/[gh] except a/abcfed/g/h', function(t) {
+  var results = glob('a/**/[gh]', {sync: true, ignore: ['a/abcfed/g/h']})
+  t.same(results, ['a/abcdef/g', 'a/abcdef/g/h', 'a/abcfed/g'])
+  t.end()
+});
+
+test('ignore, using multiple ignores', function(t) {
+  var results = glob('*', {sync: true, cwd: 'a', ignore: ['c', 'bc', 'symlink', 'abcdef']})
+  t.same(results, ['abcfed', 'b', 'cb'])
+  t.end()
+});
+
+test('ignore, using multiple ignores and globstar', function(t) {
+  var results = glob('a/**', {sync: true, ignore: ['a/c/**', 'a/bc/**', 'a/symlink/**', 'a/abcdef/**']})
+  t.same(results, ['a', 'a/abcfed', 'a/abcfed/g', 'a/abcfed/g/h', 'a/b', 'a/b/c', 'a/b/c/d', 'a/cb', 'a/cb/e', 'a/cb/e/f'])
+  t.end()
+});
+
+test('ignore a and it\'s contents', function(t) {
+  var results = glob('a/**', {sync: true, ignore: ['a/**']})
+  t.same(results, [])
+  t.end()
+});
+
+test('ignore a and it\'s contents in case of multiple globstars', function(t) {
+  var results = glob('a/**', {sync: true, ignore: ['a/**/**']})
+  t.same(results, [])
+  t.end()
+});
+
+test('ignore, get path\'s matching a/b, exclude only a/b', function(t) {
+  var results = glob('a/b/**', {sync: true, ignore: ['a/b']})
+  t.same(results, ['a/b/c', 'a/b/c/d'])
+  t.end()
+});
+
+test('ignore, get all but b', function(t) {
+  var results = glob('**', {sync: true, cwd: 'a', ignore: ['b']})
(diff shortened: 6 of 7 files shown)


UltCombo:
Odd how the only failing test is due to 2 paths out of order in Node 0.11 only.

Nice work, I'll start experimenting with this tomorrow hopefully.

//cc @contra @sindresorhus

vegetableman:
@UltCombo had those failed cases randomly thrown on my local as well but they may be due to race conditions mentioned in the README. Travis is clear now.

vegetableman:
@UltCombo did u get a chance to test it?

UltCombo:
Sorry, hadn't had time to test it throughout. I just pushed a [new branch](https://github.com/sindresorhus/globby/tree/glob-ignore) to globby using node-glob's `ignore` option in the async API.

Ignoring some files in a directory through the `ignore` option is about twice slower than globby, looks like it is because you're calling the minimatch function [inside a loop](https://github.com/isaacs/node-glob/pull/151/files#diff-d7716810d825f4b55d18727c3ccb24e6R192) which parses the pattern again in each iteration. Ideally, only a single [Minimatch class](https://github.com/isaacs/minimatch#minimatch-class) instance should be created for each `ignore` pattern. See how [globby handles it currently](https://github.com/sindresorhus/globby/blob/master/index.js#L34-L36).

Ignoring a whole dir (`dir/**`) is throwing a `RangeError: Maximum call stack size exceeded`, this might be a problem with the benchmark though.

contra:
@UltCombo thanks for taking the time to benchmark it :+1: 

So I put $100 on bountysource for this but it was on the other issue #, not sure how to move it over here but I'll contact them about that.

vegetableman:
@UltCombo Thanks for the benchmarks.

> you're calling the minimatch function inside a loop which parses the pattern again in each iteration.Ideally, only a single Minimatch class instance should be created for each ignore pattern. See how globby handles it currently.

Creating only one instance now.

> Ignoring a whole dir (dir/**) is throwing a RangeError: Maximum call stack size exceeded,

Using setImmediate to return the callback has fixed the issue. Ref:- https://github.com/logicalparadox/matcha/issues/9

And following are the results from the existing globby benchmarks:
```
matcha bench

                      negative globs (some files inside dir)
             229 op/s » globby async (working directory)
             223 op/s » globby async (upstream/master)
             241 op/s » globby sync (working directory)
             226 op/s » globby sync (upstream/master)
              59 op/s » glob-stream

                      negative globs (whole dir)
          10,075 op/s » globby async (working directory)
             368 op/s » globby async (upstream/master)
             236 op/s » globby sync (working directory)
             235 op/s » globby sync (upstream/master)
              66 op/s » glob-stream

                      multiple positive globs
              86 op/s » globby async (working directory)
              89 op/s » globby async (upstream/master)
              77 op/s » globby sync (working directory)
              74 op/s » globby sync (upstream/master)
              36 op/s » glob-stream


  Suites:  3
  Benches: 15
  Elapsed: 26,361.82 ms
```

UltCombo:
Nice!
Btw, so far I've only updated the async logic to use the `ignore` option, so you can ignore the sync results for now.

vegetableman:
Updated benchmarks on fixing globby sync:-

```
matcha bench

                      negative globs (some files inside dir)
             223 op/s » globby async (working directory)
             222 op/s » globby async (upstream/master)
             432 op/s » globby sync (working directory)
             254 op/s » globby sync (upstream/master)
              68 op/s » glob-stream

                      negative globs (whole dir)
          11,760 op/s » globby async (working directory)
             409 op/s » globby async (upstream/master)
          18,977 op/s » globby sync (working directory)
             244 op/s » globby sync (upstream/master)
              73 op/s » glob-stream

                      multiple positive globs
              98 op/s » globby async (working directory)
              96 op/s » globby async (upstream/master)
             516 op/s » globby sync (working directory)
              81 op/s » globby sync (upstream/master)
              37 op/s » glob-stream


  Suites:  3
  Benches: 15
  Elapsed: 23,045.84 ms
```

UltCombo:
Oh, I was about to start working on that, thank you very much!
Mind sending a PR to globby later?

vegetableman:
Sure. made a quick fix for now. updating it.

UltCombo:
Thanks, no hurry. Sending a quick fix plus the Matcha fix is okay, @sindresorhus and I can clean it up afterwards. `=]`

vegetableman:
Have sent fix for sync. Matcha fix?

common.js
```diff
@@ -74,6 +75,27 @@ function setopts (self, pattern, options) {
   self.statCache = options.statCache || Object.create(null)
   self.symlinks = options.symlinks || Object.create(null)
 
+  self.ignore = options.ignore || []
+
+  if(!Array.isArray(self.ignore)) {
```
isaacs: Please add space to match the style in the rest of the file.

common.js
```diff
@@ -74,6 +75,27 @@ function setopts (self, pattern, options) {
...
+    self.ignore = [self.ignore]
+  }
+  if(self.ignore.length) {
+    self.ignore = self.ignore.map(function(pattern) {
```
isaacs: This function should be abstracted out to `common.ignoreMap` or something.

common.js
```diff
@@ -74,6 +75,27 @@ function setopts (self, pattern, options) {
...
+      var endsWithGlobStar = pattern.slice(-3) === '/**',
+          endsWithStar = pattern.slice(-2) === '/*'
```
isaacs: New var each line.

common.js
```diff
@@ -74,6 +75,27 @@ function setopts (self, pattern, options) {
...
+          m = { matcher: new Minimatch(pattern) };
+
+      if(endsWithGlobStar)
+        m.gmatcher = new Minimatch(pattern.slice(0, -3));
+
+      if(endsWithStar)
+        m.smatcher = new Minimatch(pattern.slice(0, -2));
```
isaacs: What if the pattern ends with multiple `/**` parts?  Eg, `foo/**/**/**` is functionally equivalent to `foo/**`

common.js
isaacs: Oh, I see, this is already in common.  In that case, it can just be a module-private named function, if it's not needed externally.

glob.js
```diff
@@ -64,7 +65,7 @@ function glob (pattern, options, cb) {
 
   if (options.sync) {
     if (cb)
-      throw new TypeError('callback provided to sync glob')
+      throw new TypeError("callback provided to sync glob")
```
isaacs: Please don't change the quote styles.

glob.js
```diff
@@ -126,7 +127,9 @@ function Glob (pattern, options, cb) {
     cb = once(cb)
     this.on('error', cb)
     this.on('end', function (matches) {
-      cb(null, matches)
+      setImmediate(function() {
```
isaacs: Can you explain why this is added?  The `end` event should never happen synchronously.

glob.js
```diff
@@ -269,14 +272,23 @@ Glob.prototype._process = function (pattern, index, inGlobStar, cb) {
...
-    this._processReaddir(prefix, read, abs, remain, index, inGlobStar, cb)
+  //if ignored, skip _processing
+  if (isIgnored(this, abs)) {
+    return setImmediate(function() {
+      return cb();
+    });
+  }
+  else {
+    var isGlobStar = remain[0] === minimatch.GLOBSTAR
+    if (isGlobStar) {
```
isaacs: do not add `{` and `}` for one-line if/else.

glob.js
```diff
@@ -312,12 +325,14 @@ Glob.prototype._processReaddir2 = function (prefix, read, abs, remain, index, in
...
   }
 
-  //console.error('prd2', prefix, entries, remain[0]._glob, matchedEntries)
+  //console.error("prd2", prefix, entries, remain[0]._glob, matchedEntries)
 
   var len = matchedEntries.length
   // If there are no matched entries, then nothing matches.
   if (len === 0)
-    return cb()
+    return setImmediate(function() {
```
isaacs: unnecessary.

sync.js
```diff
@@ -97,13 +98,19 @@ GlobSync.prototype._process = function (pattern, index, inGlobStar) {
...
 
-  var isGlobStar = remain[0] === minimatch.GLOBSTAR
-  if (isGlobStar)
-    this._processGlobStar(prefix, read, abs, remain, index, inGlobStar)
-  else
-    this._processReaddir(prefix, read, abs, remain, index, inGlobStar)
+  //if ignored, skip processing
+  if(isIgnored(this, abs)) {
+    return
+  } else {
```
isaacs: No need to wrap this in an else, because of the return.

test/readdir-order.js
```diff
@@ -44,7 +44,9 @@ test('shuffle fs.readdir', function (t) {
 
   // make sure it works
   var d = __dirname + '/a'
-  t.notSame(fs.readdirSync(d), fs.readdirSync(d).sort())
+  process.nextTick(function() {
```
isaacs: Not relevant to this change.

isaacs:
@vegetableman The tests are thorough, and it looks like you've got the API pretty much perfect and working.  Good job!

However, there's a **lot** of stuff in this patch that is not relevant to the change.  I started commenting on it, but gave up after a while.

1. Single quotes have been changed to double.  Including in words like `isn't` in comments, even.  Don't do that.  This project uses single-quotes exclusively for strings, as of 82b8568.
2. The if/else formatting has been changed in many places.  There are two ways that this project does if blocks.  If it is a single-line, it should not use braces.  If it is a multi-line, it should use braces of the form of `} else {` all on one line.
3. Spacing around conditionals and function calls.  There should be spaces between the `if` and the opening `(` of the conditional.  Function definitions should have a space between the name and the opening `(`.
4. Many unnecessary `setImmediate` calls are added, I'm not sure why.  They aren't relevant to this feature.  If you are finding cases with strange timing errors, please provide example code and post a new issue about that.

These things make it much harder than necessary to review the change.

In general, a commit for a feature should be as small as possible, and no smaller; it should do exactly the feature that is being added, and nothing else; it should match the style of the surrounding code so well that no one can tell that the original author didn't write it themselves.

isaacs:
Oh, also, semicolons added to line endings in many places.

UltCombo:
> Have sent fix for sync. Matcha fix? Do you mean the setImmediate fix? It's part of node-glob.

Seeing as this seems to be a [Matcha issue](https://github.com/logicalparadox/matcha/issues/9) (as you've linked previously), I believe it would make sense to workaround the issue in the benchmark tests instead of the lib core.

That is, add `setImmediate` to call the benchmark test's callback instead of polluting the node-glob core with it.

vegetableman:
@isaacs Thanks for reviewing it. Have fixed the formatting and quote issues. Sorry for that. Removed ```setImmediate``` calls which were added to fix call stack errors on globby benchmarks/test issues. 

> What if the pattern ends with multiple /** parts? Eg, foo/**/**/** is functionally equivalent to foo/**

 Is it supposed to behave differently?. right now, fetching ```foo/**``` and ```foo/**/**/**``` return the same number of results.

isaacs:
Note: Travis error is spurious, fixed on aaffe249119c4640ed71e186d40516aa8e7e2d4e.

> Is it supposed to behave differently?. right now, fetching `foo/**` and `foo/**/**/**` return the same number of results.

No, it's supposed to behave the same, but the issue is that you're doing special treatment for things that end in `/**`, and stripping it off.  Why not `.replace(/(\/\*\*)+$/, '')` instead, to strip off all trailing `/**` segments?

isaacs:
Just another thought, what happens if you have this structure:

```
a
├── b/
│   └── bc/
└── c/
    └── bc/
```

And you glob for `a/**/b*` with an ignore of `a/b`.

Without the ignore, you get three matches:

```
> g.sync("a/**/b*")
[ 'a/b', 'a/b/bc', 'a/c/bc' ]
```

But with it, I'd think that only the explicit `a/b` match should be ignored.  Instead, I get:

```
> g.sync("a/**/b*",{ignore: ["a/b"]})
[ 'a/c/bc' ]
```

That seems like a bug to me.  Since there are no magic chars in the ignore (eg, it's not ignoring `a/b/**` or `a/b/*` then only the explicit `a/b` match should be omitted.

vegetableman:
> No, it's supposed to behave the same, but the issue is that you're doing special treatment for things that end in /**, and stripping it off. Why not .replace(/(\/\*\*)+$/, '') instead, to strip off all trailing /** segments?

It's fixed already. Have added a test case as well. Misunderstood you before.

>That seems like a bug to me. Since there are no magic chars in the ignore (eg, it's not ignoring a/b/** or a/b/* then only the explicit a/b match should be omitted.

Did you pull my latest changes?. Tried your scenario, it works.

vegetableman:
Merged in master.

isaacs:
@vegetableman I pulled in 4c71eaebe0e09e69d59d5a95bc9c58c882e0fdec

```
> g.sync('a/**/b*')
[ 'a/b', 'a/b/bc', 'a/c/bc' ]
> g.sync('a/**/b*', {ignore:'a/b'})
[ 'a/c/bc' ]   <--- Expect: [ 'a/b/bc', 'a/c/bc' ]
```

isaacs:
```
$ glg -n5
*   ddf2c1c vegetableman (HEAD, vegetableman/fix-ignore-branch) Merge remote-tracking branch 'origin/master' into fix-ignore-branch (6 seconds ago)
|\
| * aaffe24 isaacs (origin/master, master) make readdir-order test deterministic (48 minutes ago)
* | 4c71eae vegetableman New: Added ignore option (fixes #115) (7 hours ago)
|/
* 82b8568 isaacs Settle on single quotes (4 days ago)
* 941d53c isaacs (tag: v4.3.2, vegetableman/master) v4.3.2 (3 weeks ago)
```

vegetableman:
yep. got it. sending a fix.

vegetableman:
Fixed and added tests. Was messing about on the test directory which somehow returned correct results.

common.js
```diff
@@ -41,6 +42,29 @@ function alphasort (a, b) {
...
+
+  if (self.ignore.length) {
+    self.ignore = self.ignore.map(function (pattern) {
+      var endsWithGlobStar = pattern.slice(-3) === '/**'
+      var endsWithStar = pattern.slice(-2) === '/*'
+      var m = { matcher: new Minimatch(pattern) }
+
+      //Using indexOf to match 'a' in case of patterns like a/**/**
+      if (endsWithGlobStar)
+        m.gmatcher = new Minimatch(pattern.slice(0, pattern.indexOf('/**')))
```
isaacs: This will be incorrect in the case of `a/**/b/**` where you want it to end up with `a/**/b`, and instead, it'll be `a`.

isaacs: Also, it'll get caught up with `a/**b**/c/**`

isaacs:
Still slightly incorrect.

```
> g.sync('a/**/b*')
[ 'a/b', 'a/b/bc', 'a/c/bc' ]
> g.sync('a/b/*')
[ 'a/b/bc' ]
> g.sync('a/**/b*', {ignore:'a/b/*'})
[ 'a/c/bc' ]
```

If you're ignoring `a/b/*`, then that should *only* exclude files that are matched by `a/b/*`.  Since `a/b` isn't matched by `a/b/*`, it should not be excluded.

Also:

```
> g.sync('a/**/b*', {ignore:'a/**b/**'})
[]
```

I think all the "ends with star" or "ends with globstar" bits need to be ripped out.  It's a premature optimization.

isaacs:
The principle has to be this:

The result of `glob(pattern, {ignore:set})` is equal to to the complete subset of `glob(pattern)` not found in the union of the glob expansions of every pattern in the ignore set.

vegetableman:
@isaacs Thanks for testing. Took another crack at it, have fixed the pattern matching issues, i believe once and for all. All the above issues has been addressed.

Would be great if you consider/test these changes once, before I revert back.

> It's a premature optimization.

I don't think so. It's important because let's say i ignore node_modules folder using ```node_modules/**```, then the folder won't be read at all saving huge processing time. Please refer the above benchmark results. 

With the change:-
```
           negative globs (whole dir)
           7,233 op/s » globby async (working directory) // with the change
             409 op/s » globby async (upstream/master)
          16,077 op/s » globby sync (working directory) // with the change
             246 op/s » globby sync (upstream/master)
              71 op/s » glob-stream
```

without it:-
```
             negative globs (whole dir)
             231 op/s » globby async (working directory) // without the change
             379 op/s » globby async (upstream/master)
             428 op/s » globby sync (working directory)  // without the change
             247 op/s » globby sync (upstream/master)
              73 op/s » glob-stream
```

isaacs:
Sorry, I was unclear.  I'm not saying that the optimization is not a requirement for release of this feature.

I'm saying it's premature in advance of correct behavior.

The first step should be to implement the contract correctly, and provide a mountain of tests for every conceivable edge case. *Then* add optimizations, each time also adding more tests to guarantee that the optimization's new edge cases don't break the contracts.

I'll review again tomorrow and see if I can poke any new holes in it ;)

isaacs:
@vegetableman Thanks for your patience on this, btw.  I know it's frustrating when you make a contribution and the author puts you through the ringer, but since this module is used by many high-profile fs libs, it's important to be really careful about anything that can introduce surprising edge cases.  In the worst case, they end up depending on those bugs, and then when we fix them, it causes problems.

vegetableman:
Thanks @isaacs for your patience as well. You are awesome :+1: .

vegetableman:
Added more tests including tests for negation and brace-expansion to accelerate this PR.

isaacs:
I don't think "negative ignores" is a good idea.  Maybe we should just throw in that case.  If you want something to not be ignored, then put it in the pattern.  It opens the door for folks to do crazy stuff like `glob('**', {ignore:['!thing-i-want', '!other-thing-i-want']})` so now we're into double-negative territory.  If that's a worthwhile use-case, then let's just add multi-glob support with some sensible API surface.

Added one minor comment, will review in more detail later.

vegetableman:
Agreed. It's kind of awkward and the current functionality simply reverses an ignore and excludes other non-matching files. May be we could go the globby way of supporting multi-globs. Something like
```js
glob(['a/b', 'b/c'])
```
It's worth noting that globby supports an interesting way of handling multi-globs and corresponding ignore list. Refer:- https://github.com/sindresorhus/globby/pull/6.

UltCombo:
> If that's a worthwhile use-case, then let's just add multi-glob support with some sensible API surface.

node-glob's lack of multi-glob support is the sole reason [globby](https://github.com/sindresorhus/globby) exists.

Also worth noting that [glob-stream](https://github.com/wearefractal/glob-stream) implements the very same logic as globby as well (albeit it consumes node-glob's EventEmitter API instead of the callback/sync ones). Multi-glob support + a streaming API (#70) should make glob-stream obsolete as well.

contra:
multi-glob support and a streaming API would completely solve our issues and perf bottlenecks in gulp and I'm sure in a lot of other projects

isaacs:
Let's discuss multi-glob and streaming in other issues, plz.

isaacs:
@vegetableman Getting closer, but logic's still not quite right.

```
> g.sync('test/a/c/**')
[ 'test/a/c',
  'test/a/c/d',
  'test/a/c/d/c',
  'test/a/c/d/c/b' ]
> g.sync('test/a/c/*')
[ 'test/a/c/d' ]
> g.sync('test/a/c/**', {ignore:'test/a/c/*'})
[]
```

In this case, it should *only* ignore patterns that would be matched by `test/a/c/*`, that is, `test/a/c/d`.

However, as you can see, it's failing to even explore the child directories in `test/a/c`, because of the star handling logic.  The expected value here would be:

```
> g.sync('test/a/c/**').filter(function(p) { return p !== 'test/a/c/d' })
[ 'test/a/c',
  'test/a/c/d/c',
  'test/a/c/d/c/b' ]
```

vegetableman:
Thanks @isaacs  for finding the bug. Great catch. 
Pushed a fix with test cases. Have removed the single star check. Also, missed catching it before, since, I was passing absolute paths ( which was enabled on ```cwd```), now only relative paths are passed.

vegetableman:
@isaacs The star handing logic has been simplified with support only for globstar, that is, ignore traversing if globstar, and minimatch check for the rest. I think this could be merged.

isaacs:
Ok!  Landed this on master.

@vegetableman I had a few other suggestions, which were mostly pretty minor, and I went ahead and just made the changes myself.  Since we've gone back and forth a bunch on this, and you seemed to be getting something out of the process, I want to share the reasoning behind both:

1. The two tests were merged into a single test that just lists cases and has less repeated code.  This makes it a lot easier to add new tests, and reduces the likelihood that one day we'll add a new ignore test, and forget to update the sync test as well, etc.
2. I split up the `isIgnored` function into `isIgnored` and `childrenIgnored`, in order to remove the boolean argument.  In general, a boolean argument is often an indication that you probably are putting two different functions in the same place, and it'd be better to just have them go ahead and be two separate functions.
3. I de-inlined the argument to `ignore.map()`, since it doesn't actually need access to the closure state.

Will release with a bumped minor later today.  Might try to get a few other PRs in there as well.

vegetableman:
Thanks for the refactor @isaacs . Makes sense and is much more comprehensible.

contra:
Oooooooohhhhhhh!

Waiting on a PR for glob-stream now to incorporate this. Thanks @isaacs

vegetableman:
@contra Have sent a bounty request. Waiting for you to accept it. :money_with_wings:

contra:
@vegetableman Didn't get anything, can you try again?

contra:
@vegetableman Sent - thanks!

PR #159, opened 2015-02-06
Glob sync: trailing slash fix

### LINKED ISSUE #158: Trailing slash in pattern is matching files as well
I’ve encountered a peculiar issue and I’m not sure if it’s a bug.

I’m globbing folders with subfolders using the pattern `*/*/`. I only want to match folders, so I’m using the trailing slash to exclude files (I wish there was a `onlydir` option as well as a `nodir` option). Here’s the strange thing: The subfolder pattern is matching files as well, but when I use the pattern `*/`, no files are matched. I’ve uploaded an example for your perusal.

I’m not sure if this is a `node-glob` issue or a `minimatch` issue, but since I’ve encountered it while using `node-glob`, I put it here.

Using node 0.10.36 and node-glob 4.3.5. node-glob is `g` in the screenshot below.

image

At the moment I’m using a silly workaround to filter out files, but I’d rather have the glob pattern do that for me.

Is this a bug or intended behaviour? I did some issue searching and #49 looks tangentially related, but that was ages ago.

meyer:
This PR fixes #158.

#### `GlobSync.prototype._processSimple`

* **Line 331** Renamed `exists` to `prefixStat` so that it’d make sense when I’d refer to it later.
* **Line 367**: `abs` was being built with `path.resolve`, but `path.resolve` trims the trailing slash off files and folder paths, which means any trailing slash check on `abs` will always be negative. I switched it to `path.join` instead.
* **Lines 364-367**: The absolute path wasn’t being “absoluteified”  in a few cases, so I fixed that.

#### `GlobSync.prototype._stat`

* **Line 386**: `fs.statSync` comes undone if a file path ends in a slash but isn’t a directory, so I’m running `path.resolve` on it before `fs.statSync`.
* **Line 395-397**: `_stat` was returning false if `abs` ended with a slash but wasn’t a directory, but hey, that’s not its job, that’s `_processSimple`’s job, so I moved it up there. Also, since the trailing slash was being `resolve`d away that particular blob of code would never ever run. Poor lil guy.

Hope this all makes sense :cold_sweat: 

Ran tests with `npm test`, everything looks good.

### DIFF
--- a/sync.js
+++ b/sync.js
@@ -333,8 +333,7 @@ GlobSync.prototype._processSimple = function (prefix, index) {
   if (!this.matches[index])
     this.matches[index] = Object.create(null)
 
-  // If it doesn't exist or isn't a file or dir, then just
-  // mark the lack of results
+  // If it doesn't exist, then just mark the lack of results
   if (!prefixStat)
     return
 
@@ -377,14 +376,9 @@ GlobSync.prototype._stat = function (f) {
     if (Array.isArray(c))
       c = 'DIR'
 
-    // It exists, but not how we need it
-    if (abs.slice(-1) === '/' && c !== 'DIR')
-      return false
-
     return c
   }
 
-  var exists
   var stat = this.statCache[abs]
   if (!stat) {
     try {
--- a/sync.js
+++ b/sync.js
@@ -348,11 +348,11 @@ GlobSync.prototype._processSimple = function (prefix, index) {
     }
   }
 
-  if (prefixStat !== "DIR" && prefix.slice(-1) === "/")
-    return
-
   if (process.platform === 'win32')
     prefix = prefix.replace(/\\/g, '/')
+
+  if (prefixStat !== "DIR" && prefix.slice(-1) === "/")
+    return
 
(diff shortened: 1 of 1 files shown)


isaacs:
Thanks for getting this started!

Before this can be accepted, though, there are two things that need to be done.

First, it needs a test added (something that fails before the patch, and passes with the patch).  You can either modify an existing test to add the new case, or add a new one.   (Note that if you do add a new test, you'll have to run `npm run test-regen` on a Mac or Linux box to re-generate the bash comparison test cases.)

Second, I'm pretty sure that all of these glob.sync errors are *also* errors in the async case.  If you look through the code, you'll see that they follow one another pretty closely.  So, each of these changes in the sync flow have to have the corresponding change made in the async flow.

Let me know if you are/aren't interested in doing these bits, or if you need any help with them.  Eventually, I'm sure I or someone else will get to it if you don't find the time, but it might take a bit longer to get to.

isaacs:
Fixed by 0b729c83f15c3d78f5235f893c1ed912e6386eb7.  Thanks!

PR #166, opened 2015-02-25
`ignore` option doesn't work with the `dot` option

UltCombo:
Given glob-stream's [fixtures](https://github.com/wearefractal/glob-stream/tree/6411ea4891d24227f94f302bd52d79584af8fa9b/test/fixtures):

```
fixtures
|- .swag
`- test.coffee
```

The following test case should return an empty result set:

```js
glob('./fixtures/*swag', { ignore: './fixtures/*', dot: true }, function(err, paths) {
  console.log(paths); // should be an empty array
});
```


UltCombo:
@isaacs does this approach looks ok?

vegetableman:
@UltCombo The use case of supporting negates ```!``` on ignore is debatable. See https://github.com/isaacs/node-glob/pull/151#issuecomment-69977736. Using a negated pattern on ```ignore``` option would mean adding support for "ignoring an ignore".  In case of above test case, ```ignore: '!./fixtures/**'``` should return all matches for the pattern and not ignore them.

With regards to your fix, one way to support ```ignore```  with ```dot``` without supporting negate, is to change the line:
```js
return ignoreMap(pattern, options)
```
to 
```js
ignoreMap(pattern, Object.create(options, {nonegate: {value: true}}))
```

UltCombo:
@vegetableman my bad, I don't particularly care about negated patterns in ignore patterns, the leading `!` was a copypaste error. Though, from my understanding, the concept behind `ignore` is that anything matched by an ignore pattern should not be present in the result set, so standard syntax such as negated patterns should apply to them still I believe.

vegetableman:
Fair enough.

UltCombo:
I believe negated patterns are a rare enough use case (even more so for `ignore` patterns), but perhaps it may be useful in some use cases.

Anyway, this PR addresses mainly the `dot` option (and other options that may be passed in), so that ignore patterns follow the same rules as the main pattern.

@isaacs do you think there would be any problem passing node-glob's `options` argument directly to the Minimatch constructor as this PR does?

isaacs:
So, lemme make sure I get this right.  You're saying that the test case in the original post here should return an empty array, and as far as I can tell, it does:

```
$ tree fixtures/
fixtures/
├── .swag
└── test.coffee

0 directories, 2 files

$ node -e 'require("./glob.js")("./fixtures/*swag", { ignore: "./fixtures/**", dot: true }, console.log)'
null []
```

Is there a simple case you can provide where that doesn't work as expected?

UltCombo:
@isaacs my bad, try with a single star:

```
$ node -e 'require("./glob.js")("./fixtures/*swag", { ignore: "./fixtures/*", dot: true }, console.log)'
```

isaacs:
Aha!  This is two bugs, then!

The gstarmatcher thing is broken in this case, also:

```
$ node -e 'require("./glob.js")("./fixtures/.*", { ignore: "./fixtures/**" }, console.log)'
null []
```

In that case, I'd expect it to only be ignoring things matched by `./fixtures/**`, ie `./fixtures/test.coffee`.

It seems reasonable to me that the options provided to the top-level matching should also be passed to the ignore matcher, so this patch gets a tentative +1 from me, once I get a chance to review in a bit more detail.

Regarding the other issue, I'm not sure the best approach... dropping the optimization for `**` in an ignore seems like it's going to be problematic, since it means we have to potentially walk a huge directory tree for no reason.  On the other hand, detecting the case where this might be a problem (when `dot` isn't set and the matching pattern contains some explicit dot-portion which falls somewhere within the globstar section) is potentially tricky.  (Or maybe what I just wrote in the previous sentence is an adequate heuristic and we can go with that?)

Will think on this a little bit and set a reminder to take a look at this again very soon.

UltCombo:
Oh, I didn't know globstar does not match dot files by default until now. TIL something new.
And this is more complicated than I thought. :/

contra:
Any progress on this?

isaacs:
Can you try with the patch in #227 and see if that is less surprising?

Issue #176, opened 2015-03-09
nonull regression


schnittstabil:
Looks like there is a regression in the `nonull` handling, same as in #43:
```sh
$ node -e " console.log( require('glob').sync('xyz', {nonull: true}) ); "
[]
```
Expected:
```sh
[ 'xyz' ]
```
The behavior changed from 033402b to 3f9b970, i.e. between v4.0.6 and v4.1.0.

NB, the following works as expected:
```sh
$ node -e " console.log( require('glob').sync('xyz*', {nonull: true}) ); "
[ 'xyz*' ]
```

schnittstabil:
The async behavior changed some commits later: 5a40b67 (that's also between v4.0.6 and v4.1.0):
```sh
$ node -e " require('./glob.js')('xyz', {nonull: true}, function(e, f) { console.log(f) } ); "
[]
```

PR #177, opened 2015-03-10
Fix nonull regression

schnittstabil:
See #176


isaacs:
Good patch, good test.  Thanks!  <3

PR #182, opened 2015-03-13
Remove unnecessary .npmignore

shinnn:
Because package.json has `files` field. 9ffcb39b32d183b713013480dbc0b8b973f3fca5


shinnn:
Fixed in https://github.com/isaacs/node-glob/commit/b185987e1c8d6d192b538b194be768d0f844f40e

PR #185, opened 2015-03-15
mention realpath for match event

jldec:


isaacs:
Landed on master, thanks!

PR #186, opened 2015-03-26
Remove duplicate option description

marclipovsky:
`nonull` was listed twice in the options descriptions.


isaacs:
Fixed already by another PR.

PR #188, opened 2015-04-08
quote the versions in .travis.yml so they're matched correctly

sindresorhus:
The current one matches Node `0.1` and `0.11`.

Also test on `iojs`.


isaacs:
landed, thanks!

PR #196, opened 2015-05-21
Because we are using "files" entry in package.json

stevemao:
No need to maintain an extra metadata


shinnn:
Fixed in https://github.com/isaacs/node-glob/commit/b185987e1c8d6d192b538b194be768d0f844f40e

PR #198, opened 2015-05-23
fix an error

yanxyz:


basarat:
This is still incomplete. A better fix : https://github.com/isaacs/node-glob/pull/222

isaacs:
Landed #222 instead, thanks.

PR #218, opened 2015-09-10
Update README.md

awinograd:
Fix glob primer documentation for `?`


isaacs:
This is incorrect.  In glob patterns, `?` matches exactly 1 character.

```
$ touch fo foo fob

$ echo fo?
fob foo

$ node -p 'require("./").sync("fo?")'
[ 'fob', 'foo' ]

$ echo fo*
fo fob foo

$ node -p 'require("./").sync("fo*")'
[ 'fo', 'fob', 'foo' ]
```

PR #223, opened 2015-10-06
fix ignore with dot

gaboesquivel:
make ignore aware of the dot option
related to https://github.com/SBoudrias/mem-fs-editor/issues/45

### DIFF
--- a/common.js
+++ b/common.js
@@ -32,19 +32,20 @@ function setupIgnores (self, options) {
     self.ignore = [self.ignore]
 
   if (self.ignore.length) {
-    self.ignore = self.ignore.map(ignoreMap)
+    self.ignore = self.ignore.map(ignoreMap.bind(self))
   }
 }
 
 function ignoreMap (pattern) {
+  var options = {dot: this.dot}
   var gmatcher = null
   if (pattern.slice(-3) === '/**') {
     var gpattern = pattern.replace(/(\/\*\*)+$/, '')
-    gmatcher = new Minimatch(gpattern)
+    gmatcher = new Minimatch(gpattern, options)
   }
 
   return {
-    matcher: new Minimatch(pattern),
+    matcher: new Minimatch(pattern, options),
     gmatcher: gmatcher
   }
 }
--- a/test/00-setup.js
+++ b/test/00-setup.js
@@ -13,6 +13,7 @@ var fixtureDir = path.resolve(__dirname, 'fixtures')
 
 var files =
 [ "a/.abcdef/x/y/z/a"
+, "a/.abcdef/x/y/z/b"
 , "a/abcdef/g/h"
 , "a/abcfed/g/h"
 , "a/b/c/d"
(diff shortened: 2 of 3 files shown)


sindresorhus:
:+1:

isaacs:
This is a bit more complicated of a situation than it might appear.  (More conversation at https://github.com/isaacs/node-glob/pull/166 that might be interesting to review.)

I'd like to see a lot more testing and thought around the cases here:

1. `dot: false`, search for `a/**`, ignore `a/x/**`, file at `a/x/.y/z` exists.  Should `a/x/.y/z` be included in matches?
2. `dot: true`, search for `a/**`, ignore `a/x/**`, file at `a/x/.y/z` exists.  Should `a/x/.y/z` be included in matches?
3. `dot: false`, search for `a/**`, ignore `a/**/z`, file at `a/x/.y/z` exists.  Should `a/x/.y/z` be included in matches?
4. `dot: true`, search for `a/**`, ignore `a/**/z`, file at `a/x/.y/z` exists.  Should `a/x/.y/z` be included in matches?

If the answer to (1) is "no", then we should set `dot: true` for the gsmatcher always.  If the answer is "yes", then this patch seems appropriate, but then the optimization around ignoring `**` needs to be removed, which will make things a lot more complicated.

The answer to (2) and (4) seems to be "no" always, so this patch looks correct from that pov.

The really tricky bit is (3), and I'm not sure I have a good answer here.  People seem to be confused either way.  It might be acceptable to just say, look, for `ignore` patterns, they match dot files, that's just how it is, as long as that's documented clearly.  Another approach might be to say that it follows the same `dot` setting as the overall pattern, which is what this patch does.

To be honest, I've been procrastinating making a decision on this because I'm not sure what the right call is, and it's probably a breaking change either way.

What're your thoughts on this, as a person using glob in your program?

isaacs:
Other cases:

- `dot: false`, search for `a/**/.y/z`, ignore `a/x/**`
- `dot: true`, search for `a/**/.y/z`, ignore `a/x/**`

Really, we need a comprehensive matrix of test cases and a clear decision on each one.

gaboesquivel:
IMO `dot` should be an overall pattern.. it easier to reason about this if we do so.
You explicitly tell to node-glob to ignore dotfiles or not through the dot option.
And if you set `dot=true`, you are able to ignore dotfiles through the `ignore` option 

based on this [case 3](https://github.com/isaacs/node-glob/pull/223#issuecomment-150309340) would be always __no__, since `dot=false`

I'll work on a more comprehensive matrix of tests

isaacs:
Another interesting case:

`dot: false`, search for `a/**/.y`, ignore `a/x/**`.  Files at `a/x/.y` and `a/z/.y` both exist.  What gets returned?  Does it ignore `a/x/.y`, or not?  (Note that `a/x/.y` does not match `a/x/**` if `dot` is false, but it does match `a/**/.y`, because of the explicit dot.)

The more I think about this, the more I think that `dot:true` should be set for ignore patterns always.  It should be documented that way and just take the major version hit.

isaacs:
Feel free to mess around with the patch in #227 and let me know if it is more or less surprising in practice.

gaboesquivel:
Thanks @isaacs. It seems that always dot=true in ignores is the way to go. 
I'm closing this in favor of #227

Issue #207, opened 2015-07-10
Issue with documentation: "ignore" option


SergioMorchon:
On the root README.md, options section, it says:
ignore Add a pattern or an array of patterns to exclude matches.

But coding `{ignore: /wfl\.d\.ts$/}` breaks with:
`
TypeError: Object /wfl\.d\.ts$/ has no method 'slice'
    at ignoreMap (C:\Users\smorchon\node_modules\glob\common.js:59:15)
    at Array.map (native)
    at setupIgnores (C:\Users\smorchon\node_modules\glob\common.js:53:31)
    at setopts (C:\Users\smorchon\node_modules\glob\common.js:102:3)
    at new Glob (C:\Users\smorchon\node_modules\glob\glob.js:116:3)
    at glob (C:\Users\smorchon\node_modules\glob\glob.js:73:10)
`

SergioMorchon:
It looks like "ignore" option must be a string, and not a RegExp object. I think it would be fine if would be able to work with RegExp objects, as they are more readables and flexibles.

Or update the README.md specifying tha ignore must be string typed.

isaacs:
The `ignore` option is a glob pattern string.  To replicate the RegExp you're providing, it'd be `"*wfl.d.ts"`

doc patch welcome.

SergioMorchon:
Oh, sorry. I misunderstood the explanation. When I saw "pattern" I inmediatly thinked of RegExp, not glob... Sounds stupid.

Sorry!

You can destroy this issue with greek fire if you wanna to.

oncletom:
Got stuck too while to exclude `node_modules` as well. I first tried with `/node_modules/` and figured out with @isaacs comment I had to type `'node_modules/**'`.

Thanks folks!

PR #225, opened 2015-10-10
Be more explicit about which kind of pattern is supported by the `ign…

oncletom:
…ore` option.

Cf. #207 comment (got stuck into it as well).


isaacs:
Landed on master, thanks.

oncletom:
:+1:

Issue #226, opened 2015-10-16
dot: true for ignore patterns


jaxesn:
Should the dot param be passed through to minimatch when creating the ignore matcher? 

If paths.root contains a folder that starts with a dot, the spec.coffee files are not ignored.

`glob(paths.root + "/**/*.coffee", {ignore: '**/*_spec.coffee', dot: true}, next)`

PR #227, opened 2015-10-25
Ignores are always dot:true

isaacs:
Related issues:

- #223
- #226
- #166 

[EDIT: Added docs]


gaboesquivel:
:+1:

sindresorhus:
:+1:

isaacs:
If anyone wants to weigh in on the substantial change here, take a look at these test cases: https://github.com/isaacs/node-glob/blob/dot-ignore/test/ignore.js#L34-L39

The relevant file structure looks like this:

```
$ tree a/{x,z}
a/x
└── .y/
    └── b
a/z
└── .y/
    └── b

2 directories, 2 files
```

The linked test cases go `[pattern, ignore, results]`.  So, eg, a pattern of `a/**/b`, with an ignore set to `a/x/**` will prevent `a/x/.y/b` from matching, even though the `**` would not normally match the `.y` path portion.

aredridel:
This sure seems like simpler behavior

isaacs:
Alright, the requisite time for objection and testing has passed, and it seems to make everyone's use cases better.  I'm gonna go ahead and land this.  It's a SemVer-major bump, because of the change in behavior, but it's the right change, I believe.

UltCombo:
This does fix a bunch of issues and is indeed less surprising.

Though, I don't understand why the other rules for the ignore patterns wouldn't match the main matching pattern's rules? (#166)

Perhaps it is better to keep this simple, but having different rules for the matching and ignore patterns will likely lead to unexpected/counter-intuitive behavior.

isaacs:
@UltCombo I don't understand the question.

I think experience thus far has shown that having ignore patterns not match dot files is always more confusing than the alternative, even when the main pattern doesn't have `dot: true`.

UltCombo:
@isaacs I agree that ignore patterns being always `dot: true` makes sense. I was talking about other Minimatch options, such as `nocase`, `noglobstar`, `matchBase`. Shouldn't the Minimatch instances for the ignore patterns have these options set to the same value as the matching glob pattern?

thirdcreed:
node -p 'require("glob").sync("/home/thirdcreed/Projects/glob-stream/test/fixtures/*swag",{ignore:["/home/thirdcreed/Projects/glob-stream/test/fixtures/**"],dot:true,cwd: "/home/thirdcreed/Projects/glob-stream/test",cwdbase:false,nonull:false})'

This works for me in milliseconds, but [this code](https://github.com/thirdcreed/glob-stream/blob/master/index.js#L26) times out  on  only [one test](https://github.com/gulpjs/glob-stream/blob/master/test/main.js#L220), with identical  parameters.  An end event is never emitted.

Any insight you can give me into what I might be doing wrong would be much appreciated.

PR #229, opened 2015-11-11
fix readme.md

yanxyz:
`<String>` should be escaped in Markdown.


isaacs:
I decided to just `<code>` them all with backticks. Thanks for the nudge :)

PR #239, opened 2015-12-28
Breaking change in 6.0.2

denis-sokolov:
eb6319ff5 broke backward compatibility for me.
I do not yet have a simple example to take a look at, but I have isolated the breakage to this single file change.

From looking at the commit the change seems surely functional - the option defaults have been changed!

### DIFF
--- a/common.js
+++ b/common.js
@@ -104,6 +104,11 @@ function setopts (self, pattern, options) {
 
   self.nomount = !!options.nomount
 
+  // disable comments and negation unless the user explicitly
+  // passes in false as the option.
+  options.nonegate = true
+  options.nocomment = true
+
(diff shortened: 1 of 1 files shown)


denis-sokolov:
Confirmed breakage. `minimatch` documentation on the `nonegate` flag:

> This is perhaps relevant if you wish to start the pattern with a negative extglob pattern like !(a|B)

This is exactly the case for me, the simplified pattern is `!(node_modules|typings)/**/*.ts`

isaacs:
Indeed!  Landed this with a comment so that I don't go breaking it again.  Thanks for the patch.

denis-sokolov:
You could add a test so that you don’t go breaking it again!

PR #242, opened 2016-01-14
Fix follow behaviour

medains:
I noticed bad symlink following behaviour from eslint usage which results in scanning files within symlinked directories even with follow:false set.

This PR fixes async and sync behaviour, adds tests for both in the specific eslint use case and deals with a bash oddity:  whilst ** doesn't follow a symlink **/* will.  This may be an edge-case bug in the bash implementation too - but I don't want to follow this rabbit-hole any deeper right now.

### DIFF
--- a/test/match-base.js
+++ b/test/match-base.js
@@ -11,7 +11,7 @@ var expect = [
 ]
 
 if (process.platform !== 'win32')
-  expect.push('a/symlink/a', 'a/symlink/a/b/c/a')
+  expect.push('a/symlink/a')
 
 t.test('chdir', function (t) {
   var origCwd = process.cwd()
--- a/glob.js
+++ b/glob.js
@@ -490,7 +490,7 @@ Glob.prototype._readdirInGlobStar = function (abs, cb) {
       self.cache[abs] = 'FILE'
       cb()
     } else
-      if (!isSym || this.follow ) {
+      if (!isSym || this.follow) {
         self._readdir(abs, false, cb)
       } else {
         cb()
--- a/glob.js
+++ b/glob.js
@@ -490,7 +490,11 @@ Glob.prototype._readdirInGlobStar = function (abs, cb) {
       self.cache[abs] = 'FILE'
       cb()
     } else
-      self._readdir(abs, false, cb)
+      if (!isSym || this.follow ) {
+        self._readdir(abs, false, cb)
+      } else {
+        cb()
+      }
   }
 }
 
--- a/sync.js
+++ b/sync.js
@@ -250,8 +250,13 @@ GlobSync.prototype._readdirInGlobStar = function (abs) {
   // don't bother doing a readdir in that case.
   if (!isSym && !lstat.isDirectory())
     this.cache[abs] = 'FILE'
-  else
-    entries = this._readdir(abs, false)
+  else {
+    if (!isSym || this.follow) {
+        entries = this._readdir(abs, false)
+    } else {
+        entries = [];
+    }
+  }
 
   return entries
 }
@@ -340,15 +345,16 @@ GlobSync.prototype._processGlobStar = function (prefix, read, abs, remain, index
   var gspref = prefix ? [ prefix ] : []
   var noGlobStar = gspref.concat(remainWithoutGlobStar)
 
-  // the noGlobStar pattern exits the inGlobStar state
-  this._process(noGlobStar, index, false)
-
   var len = entries.length
   var isSym = this.symlinks[abs]
 
   // If it's a symlink, and we're in a globstar, then stop
   if (isSym && inGlobStar)
     return
+
+  // the noGlobStar pattern exits the inGlobStar state
+  this._process(noGlobStar, index, false)
+
 
   for (var i = 0; i < len; i++) {
     var e = entries[i]
--- a/test/00-setup.js
+++ b/test/00-setup.js
@@ -102,7 +102,7 @@ var globs =
   ,"{./*/*,/tmp/glob-test/*}"
   ,"{/tmp/glob-test/*,*}" // evil owl face!  how you taunt me!
   ,"a/!(symlink)/**"
-  ,"a/symlink/a/**/*"
+  ,"a/symlink/a/**"
   ]
 var bashOutput = {}
 var fs = require("fs")
--- a/test/bash-results.json
+++ b/test/bash-results.json
@@ -132,9 +132,9 @@
     "a/x",
     "a/z"
   ],
-  "a/symlink/a/**/*": [
+  "a/symlink/a/**": [
+    "a/symlink/a",
     "a/symlink/a/b",
-    "a/symlink/a/b/c",
-    "a/symlink/a/b/c/a"
+    "a/symlink/a/b/c"
   ]
 }
--- a/test/ok-symlink.js
+++ b/test/ok-symlink.js
@@ -23,9 +23,9 @@ test('set up working symlink', function (t) {
   fs.symlinkSync('../working-target', 'a/working-link/link')
   t.end()
 })
-/*
+
 test('async follow test', function (t) {
-  var pattern = 'a/working-* /** /*.js';
+  var pattern = 'a/working-*/**/*.js';
   var expected = [
     'a/working-target/file.js',
     'a/working-link/link/file.js',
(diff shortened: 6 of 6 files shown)


medains:
@isaacs : I assume that this is not so controversial, since there's been no comment on it.  Happy for this to be included in the next version?

julienw:
`**` actually follows symlinks for me when I use `follow: true`...

medains:
@julienw : The issue is what happens when follow: false is set

julienw:
OK thanks. sorry for the noise :)

test/match-base.js
```diff
@@ -11,7 +11,7 @@ var expect = [
 ]
 
 if (process.platform !== 'win32')
-  expect.push('a/symlink/a', 'a/symlink/a/b/c/a')
+  expect.push('a/symlink/a')
```
isaacs: This is a breaking change, and a deviation from Bash 4.3.

```
$ cd test/fixtures/

$ echo **/a*
a a/abcdef a/abcfed a/symlink/a a/symlink/a/b/c/a

$ echo $BASH_VERSION
4.3.30(6)-release
```

The `**` should be followed through to `a/symlink/a/b/c/a`, but then it should refuse to go any further subdirectories *past* the symlink.

isaacs:
Sorry for the delay responding to this.

Unfortunately, there's a bug here.  Instead of removing the test fixture and replacing it with something else, just add the new one.  Apply this diff:

```patch
diff --git a/test/00-setup.js b/test/00-setup.js
index 541e4a1..e48e2cc 100644
--- a/test/00-setup.js
+++ b/test/00-setup.js
@@ -102,6 +102,7 @@ var globs =
   ,"{./*/*,/tmp/glob-test/*}"
   ,"{/tmp/glob-test/*,*}" // evil owl face!  how you taunt me!
   ,"a/!(symlink)/**"
+  ,"a/symlink/a/**/*"
   ,"a/symlink/a/**"
   ]
 var bashOutput = {}
diff --git a/test/bash-results.json b/test/bash-results.json
index b4a1ffc..aafd3e9 100644
--- a/test/bash-results.json
+++ b/test/bash-results.json
@@ -132,6 +132,11 @@
     "a/x",
     "a/z"
   ],
+  "a/symlink/a/**/*": [
+    "a/symlink/a/b",
+    "a/symlink/a/b/c",
+    "a/symlink/a/b/c/a"
+  ],
   "a/symlink/a/**": [
     "a/symlink/a",
     "a/symlink/a/b",
```

(You can also get this by just editing the 00-setup.js and running `npm run test-regen`.)

With that, there's a failing test:

```
$ tap test/bash-comparison.js
test/bash-comparison.js ........................... 584/586
  a/symlink/a/**/*
  not ok a/symlink/a/**/*
    +++ found
    --- wanted
     [
       "a/symlink/a/b"
       "a/symlink/a/b/c"
    -  "a/symlink/a/b/c/a"
     ]
    at:
      line: 50
      column: 9
      file: test/bash-comparison.js
    stack: |
      test/bash-comparison.js:50:9
      f (node_modules/once/once.js:17:25)
      Glob.<anonymous> (glob.js:146:7)
      Glob._finish (glob.js:185:8)
      done (glob.js:172:12)
      Glob._processGlobStar2 (glob.js:625:12)
      glob.js:614:10
      RES (node_modules/inflight/inflight.js:23:14)
      f (node_modules/once/once.js:17:25)
      lstatcb_ (glob.js:509:9)
      RES (node_modules/inflight/inflight.js:23:14)
      f (node_modules/once/once.js:17:25)
      FSReqWrap.oncomplete (fs.js:82:15)

  a/symlink/a/**/* sync
  not ok should match shell (sync)
    +++ found
    --- wanted
     [
       "a/symlink/a/b"
       "a/symlink/a/b/c"
    -  "a/symlink/a/b/c/a"
     ]
    at:
      line: 61
      column: 7
      file: test/bash-comparison.js
    stack: |
      test/bash-comparison.js:61:7
      test/bash-comparison.js:54:9
      f (node_modules/once/once.js:17:25)
      Glob.<anonymous> (glob.js:146:7)
      Glob._finish (glob.js:185:8)
      done (glob.js:172:12)
      Glob._processGlobStar2 (glob.js:625:12)
      glob.js:614:10
      RES (node_modules/inflight/inflight.js:23:14)
      f (node_modules/once/once.js:17:25)
      lstatcb_ (glob.js:509:9)
      RES (node_modules/inflight/inflight.js:23:14)
      f (node_modules/once/once.js:17:25)
      FSReqWrap.oncomplete (fs.js:82:15)

total ............................................. 584/586


  584 passing (733.131ms)
  2 failing
```

If I can work out what's going wrong in the little while this evening, I'll fix and land this.  If you've got any ideas, that's welcome too :)

isaacs:
Ah, I see the problem.

> This PR fixes async and sync behaviour, adds tests for both in the specific eslint use case and deals with a bash oddity: whilst ** doesn't follow a symlink */ will. This may be an edge-case bug in the bash implementation too - but I don't want to follow this rabbit-hole any deeper right now.

Unfortunately, you'll find that this module exists at the bottom of a rather deep rabbit hole :)

By definition, from the point of view of this module, Bash 4.3's glob behavior is correct.  That's the reference implementation that this module intends to follow 100%, even when that behavior is arguably "weird".

By default, each of the optional flags are either true, or default behavior.  So, `follow: false` doesn't mean "never follow symbolic links".  It means, "only follow symbolic links in the cases where Bash 4.3 follows symbolic links."  `follow: true`, on the other hand, means "follow all symbolic links always" (which, incidentally is how Bash 4.1 handled symlinks encountered in globstar traversals, and how this module used to do it until Bash 4.3 changed its behavior, and I locked the reference implementation version that this module targets in JavaScript.)

To deviate from Bash 4.3 behavior, we'd have to add a flag which is off by default, but we've already got a lot of flags, and adding a `nofollow: true` reminds me of spammy email opt-in/out checkboxes ("don't not uncheck this box unless you don't want to stay unsubscribed to our newsletter").  Plus, what do you do with `{follow:true,nofollow:true}`?

We *could* conceivably say that `follow:false` means "no symlinks ever", and `follow: true` means "follow all symlinks", and `follow: null/undefined` means "do what Bash 4.3 does".  I don't love that either, because now if you just try to un-set a config by setting `globOpt.follow = false`, you can end up deviating from Bash behavior.  In my experience, the true/false/null option always just ends up getting confusing, and it's best to just keep it to 2 options with a boolean cast.

So, any of these are going to be breaking changes, and additional complexity.

I'm sorry for what probably seems like unnecessarily pedantic adherence to an arbitrary code contract.  Because a lot of popular command line tools rely on this module, I've learned to be very careful about any semantic changes.

julienw:
>  I don't love that either, because now if you just try to un-set a config by setting globOpt.follow = false

But you could use `globOpt.follow = null` or even `delete globOpt.follow` in that case ? The latter is the closest to "unsetting a config" :)

For context and rationale, I think @medains is ultimately trying to change a behavior in eslint (see the link at the top of this PR).

isaacs:
> But you could use `globOpt.follow = null`

Yes, that is what one would have to do.  But that is still a very subtle breaking change.

It's fine to say that this is "unexpected behavior", but it's behaving exactly like the reference implementation it's intending to emulate.  So... unexpected by whom?  If `false` and `null` start having different effects on a field that looks like a boolean flag, then that's also pretty unexpected.

Maybe the answer is a `--no-glob` option for eslint?

Overloading the `follow` flag is not the right call here.

julienw:
> Maybe the answer is a --no-glob option for eslint?

I don't follow you :) How could eslint have a different behavior if the support is not in node-glob  (except reimplementing this -- which it did in the past IIRC) ?

isaacs:
Glob never had a "do not follow any symlinks" option. `follow` was added to force it to follow symlinks in globstar traversals, unlike bash. 

Esliny could have an option to just not use glob at all. Then you wouldn't have to worry about following symbolic links. But maybe that won't work either?

I'm not sure the best answer for your use case. I know that deviating from bash behavior in this module is a bad idea, though, and pseudo-boolean-three-state option flags are flaky and error prone. 

You could also turn on `realpath:true` in the glob options for eslint. Then you'll always get the full real path, and there will never be duplicates that refer to the same actual file.

medains:
Thanks for going over this issue - I'll put something in for eslint around realpath:true

PR #243, opened 2016-01-16
Added the Promise API to the Glob prototype

thaggie:
I've added the [Promise API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Methods_2) to `Glob` objects, this is useful for keeping the indentation flat when doing further I/O operations on the returned matches.

```
glob('**/*.js')
.then(function(matches) {
  // do something
})
.catch(function(error) {
  // handle the error
})
```

### DIFF
--- a/test/promise.js
+++ b/test/promise.js
@@ -58,7 +58,7 @@ test('catch fulfilled promise - catch/then/then', function(t) {
 
 test('rejected promise - catch/then', function(t) {
   
-  expectError(t)
+  expectError()
 
   glob('*')
   .catch(function(err) {
@@ -74,7 +74,7 @@ test('rejected promise - catch/then', function(t) {
 
 test('rejected promise - then/catch/then', function(t) {
 
-  expectError(t)
+  expectError()
 
   glob('*')
   .then(function(res) {


isaacs:
I'm not opposed to this in principle, however:

1. This should be a standalone module that can promisify a thing given an EventEmitter and a "resolution" event (with "error" considered the "rejected" event, of course).  Flow-control boilerplate doesn't belong in this module.
2. That module should work on Nodes less than 0.12 (I'm told Bluebird is the thing all the cool kids use these days for their cross-platform promises.)  That's what's breaking the Travis-CI test.
3. It seems like this isn't matching the Promises/A+ spec, which says that `.catch()` should behave quite a bit differently from how it's implemented here.

Another approach would be, rather than tacking `.then()` and `.catch()` onto the Glob prototype, just create a `glob.promise(pattern, [options]) => Promise` method, since that's really what you'd probably want to use anyway, and that's pretty trivial.  Might look something like this:

```javascript
glob.promise = function (pattern, options) {
  return new Promise(function (resolve, reject) {
    var g = new Glob(pattern, options)
    g.once('end', resolve)
    g.once('error', reject)
  })
}
```

thaggie:
The `glob.promise` route is a little tedious (for consumers) but probably the cleanest one.

Personally I prefer things not to drag in dependencies and so I'd prefer that when you use a system without promises you just don't get promises - this is the way I implemented the original but not for the tests as I had naively assumed the CI would be just using a modern system.

isaacs:
Another route is just to promise up the (lower case) `glob()` function of a callback isn't provided, but if that returns a Glob instance today, it'd be a breaking change.

popomore:
Maybe you can use  https://github.com/node-modules/mz-modules

PR #253, opened 2016-03-09
Add `followsymlinkscyles` option

Y--:
Default to `true`

If false, won’t follow multiple times the same cycle through symbolic links

### DIFF
--- a/common.js
+++ b/common.js
@@ -7,6 +7,7 @@ exports.finish = finish
 exports.mark = mark
 exports.isIgnored = isIgnored
 exports.childrenIgnored = childrenIgnored
+exports.getPathDepth = getPathDepth
 
 function ownProp (obj, field) {
   return Object.prototype.hasOwnProperty.call(obj, field)
@@ -68,6 +69,7 @@ function setopts (self, pattern, options) {
   self.realpath = !!options.realpath
   self.realpathCache = options.realpathCache || Object.create(null)
   self.follow = !!options.follow
+  self.followsymlinkscyles = options.followsymlinkscyles !== false
   self.dot = !!options.dot
   self.mark = !!options.mark
   self.nodir = !!options.nodir
@@ -85,6 +87,7 @@ function setopts (self, pattern, options) {
   self.cache = options.cache || Object.create(null)
   self.statCache = options.statCache || Object.create(null)
   self.symlinks = options.symlinks || Object.create(null)
+  self.followed = options.followed || Object.create(null)
 
   setupIgnores(self, options)
 
@@ -112,6 +115,12 @@ function setopts (self, pattern, options) {
 
   self.minimatch = new Minimatch(pattern, options)
   self.options = self.minimatch.options
+}
+
+var pathSepRE = new RegExp(path.sep === '\\' ? '\\\\' : path.sep, "g");
+function getPathDepth(p) {
+  var m = p.match(pathSepRE)
+  return m && m.length || 0;
 }
 
 function finish (self) {
--- a/glob.js
+++ b/glob.js
@@ -489,11 +489,30 @@ Glob.prototype._readdirInGlobStar = function (abs, cb) {
 
   // follow all symlinked directories forever
   // just proceed as if this is a non-globstar situation
-  if (this.follow)
-    return this._readdir(abs, false, cb)
+  var self = this
+  if (this.follow) {
+    if (this.followsymlinkscyles) {
+      return this._readdir(abs, false, cb);
+    }
+
+    return fs.realpath(abs, this.realpathCache, function(err, absRealPath) {
+      if (err) { return cb(); } // doesn't exist : ignore
+
+      if (!ownProp(self.followed, absRealPath)) {
+        self.followed[absRealPath] = abs;
+        return self._readdir(abs, false, cb);
+      } else if (self.followed[absRealPath] === abs) {
+        return self._readdir(abs, false, cb);
+      } else if (common.getPathDepth(abs) < common.getPathDepth(self.followed[absRealPath])) {
+        self.followed[absRealPath] = abs;
+        return self._readdir(abs, false, cb);
+      } else {
+        return cb();
+      }
+    });
+  }
 
   var lstatkey = 'lstat\0' + abs
-  var self = this
   var lstatcb = inflight(lstatkey, lstatcb_)
 
   if (lstatcb)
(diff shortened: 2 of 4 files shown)


isaacs:
How is this different from `follow: true`?

Can you explain the use case in more detail?  And, is there a way to phrase the option so that it doesn't default to `true`?  Most of the other flags default to false, and I'd prefer to stick to that model.

Y--:
Hi @isaacs, thanks for having a look at this PR and sorry I should indeed have included more details.

`follow : true` will follow the cycles indefinitely with the `**/*` pattern until eventually an error occurs such as the path is too large, or an "ELOOP" error is thrown.

Consider the following snippet, that would run in a test script at the root of the `glob` project, with the fixtures generated by the tests :
```
var Glob    = require('./glob').Glob;
var pattern = 'test/fixtures/a/**/*';
var options = {
  follow : true,
  nodir  : true,
  sync   : true
};

console.time('With followsymlinkscyles = true (default)');
var mg = new Glob(pattern, options)
console.timeEnd('With followsymlinkscyles = true (default)');

console.time('With followsymlinkscyles = false');
options.followsymlinkscyles = false;
var mg = new Glob(pattern, options)
console.timeEnd('With followsymlinkscyles = false');
```

On my side I get something like : 
```
With followsymlinkscyles = true (default): 37.691ms
With followsymlinkscyles = false: 5.951ms
```

In the first case, the recursion stops because of the Error :
`Error: ELOOP: too many symbolic links encountered, stat 'test/fixtures/a/symlink/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c/a/b/c'`

But this can take a lot of time before this error is thrown.

In the real-world scenario I use, (within the Karma framework), since I have a lot of nested links between my repositories, the program run for several minutes before it eventually crashes with an out of memory error.

This option `followsymlinkscyles` stops the recursion when we already traversed one time entierly down the loop.

The fix might very well be naive and I'm totally open to discuss an other solution, but that was the first one that came to me.

Of course I'm totally open regarding the name. 
I thought that the name was more explicit this way and I saw the `strict` option being `true` by default so I assumed it was ok :
https://github.com/isaacs/node-glob/blob/5a40b676a16fbf6a8799977dc86fbc61e40c739f/common.js#L60
Would `dontfollowsymlinkscycles` work for you? (it looks a bit long though...)

Cheers,

Yves

isaacs:
So, you want to follow symbolic links, but _not_ symbolic links that have already been traversed, is that correct?

If I'm understanding the use-case properly, you can do get almost this behavior today by setting `realpath: true` in the options.  (Then you're guaranteed to only get each path once.)

```
> require('./').sync('test/fixtures/a/symlink/**')
[ 'test/fixtures/a/symlink',
  'test/fixtures/a/symlink/a',
  'test/fixtures/a/symlink/a/b',
  'test/fixtures/a/symlink/a/b/c' ]
> require('./').sync('test/fixtures/a/symlink/**', { follow: true }).length
100
> require('./').sync('test/fixtures/a/symlink/**', { follow: true, realpath: true })
[ '/Users/isaacs/dev/js/glob/test/fixtures/a/symlink',
  '/Users/isaacs/dev/js/glob/test/fixtures/a/symlink/a',
  '/Users/isaacs/dev/js/glob/test/fixtures/a/symlink/a/b' ]
```

Of course, the issue is that setting `{ follow: true, realpath: true }` is filling up with duplicate symbolic links and then filtering them down at the very end.  It'd be much better for performance to call `fs.realpath` along the way instead, and avoid traversing directories multiple times if `realpath: true` is set, since they won't end up in the result set anyway.  (This would also fix the weird issue that `realpath: true` glob objects still emit `match` events with the non-realpath paths, which would be a breaking change, but an improvement, imo.)

If you want output that lists the symbolic links, but and want to follow those symbolic links, and *don't* want them to be realpath'ed in the result set, then that's going to have to be a different option.  But I think there's a cheaper way to do that than calling `fs.realpath`.

> strict option being true by default

Historical accident :)  Once upon a time, it was false by default.

isaacs:
Note that, if you don't traverse any directories that have been seen already (or worse, resolve them to a realpath along the way), then you can run into weird situations.

For example, if you have `foo/a/b/c` and `bar @-> foo`, then the pattern `{foo/a,bar/*/b/*}` would match on `foo/a` and then would have to know that it SHOULD traverse `bar` (and re-traverse bar/a) to find `bar/a/b/c`, because that path doesn't match the pattern if we're calling `bar` by its realpath name, `foo`.

Y--:
Sorry for the late reply, I didn't get any time to work on this lately.
I've tried the "realpath" option and since as you pointed it, it does traverse all the symlink before filtering them at the end, this solution isn't viable. 
When I use it on my karma tests, it ends with nice messages like :

```
<--- Last few GCs --->

  119806 ms: Scavenge 1396.7 (1456.2) -> 1396.7 (1456.2) MB, 0.3 / 0 ms (+ 103.5 ms in 1 steps since last GC) [allocation failure] [incremental marking delaying mark-sweep].
  120805 ms: Mark-sweep 1396.7 (1456.2) -> 1396.7 (1456.2) MB, 999.4 / 0 ms (+ 192.6 ms in 11 steps since start of marking, biggest step 103.5 ms) [last resort gc].
  121820 ms: Mark-sweep 1396.7 (1456.2) -> 1396.7 (1456.2) MB, 1014.7 / 0 ms [last resort gc].


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x1c27036e3ac1 <JS Object>
    1: _stat [node-glob/sync.js:479] [pc=0x14ef96c5c1a7] (this=0x1c27036fd551 <a GlobSync with map 0x1321a6b47a19>,f=0x3e43b5004101 <String[621]

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory
```

Also I does agree with you that this option might lead to weird situations, but I'm not sure what would be a better solution to solve my issue then?

floribon:
+1, running out of memory using symlinks in my packages

boutils:
+1 for @Y-- , can we please merge this PR?

isaacs:
No longer an issue in v9.

PR #255, opened 2016-03-17
Simplify extend function

mina86:
Instead of creatin a temporary array of all they keys of the ‘add’
object, simply iterate over them using for-in construct instead.
This makes the code shorter and, as hinted above, gets rid of an
unnecessary array.  The patch also changes the function to have
single return which is also nice.

### DIFF
--- a/glob.js
+++ b/glob.js
@@ -81,14 +81,10 @@ var GlobSync = glob.GlobSync = globSync.GlobSync
 glob.glob = glob
 
 function extend (origin, add) {
-  if (add === null || typeof add !== 'object') {
-    return origin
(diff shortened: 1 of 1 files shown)


isaacs:
Am I correct in thinking this has no functional impact?

This way is fine, of course, but I usually don't accept patches that don't make some kind of measurable change.

mina86:
O_o
I guess increased readability is indeed not a measurable change. But here are some measurable changes:
* code is now four lines shorter,
* Object.keys creates array for no reason,
* Object.keys is ECMAScript 5.1.

PR #260, opened 2016-05-19
gitignore syntax for ignore option

gronke:
- gitignore spec syntax for ignoring files
- breaks unit test expectations in `test/ignore.js` (tests updated accordingly)


gronke:
This draft removed the minmatch syntax for demonstration purposes. If we consider gitignore syntax I suggest we add an option to switch between minmatch and gitignore style (that for compatibility reasons defaults to minmatch).

isaacs:
This would need to be reconsidered in v9.  More likely I'll go with a more flexible function based approach, which can then be easily extended to any ignoring semantics you want to implement.

Issue #173, opened 2015-03-07
Allow setup path separator of result


rentalhost:
Ok. glob() not accept match with back-slash. But the result need match with system path.sep, or at least, allow we set it manually. Else, in some cases, we need replace it "manually".

My suggestion is an option to set path separator like `{ "sep": path.sep }` or `{ "sep": "/$" }`.

isaacs:
I would be open to a change to support back-slashes on windows machines.  Probably the best solution would be to, on windows, replace all back-slashes with forward-slashes, and then swap back to back-slashes in the result.

Part of the blocker to doing stuff for windows is that I don't have a windows environment for CI tests, and it's quite tedious to run tests in Windows since I develop in a unix environment.  Do you have any tips on this?

rentalhost:
I did not understand this part:

> replace all back-slashes with forward-slashes, and then swap back to back-slashes in the result.

You mean, on unit test, you will replace `\` to `/` to match with result, is that? If yes, is a good idea.

I opened a topic on SO to listen someone about test unit in all systems: http://stackoverflow.com/questions/28918855/running-node-tests-in-all-systems-without-have-it

isaacs:
No, I mean, at the very start, we replace `\` with `/`, so that the logic all works as it is.  Then, before emitting a `match` event or calling the cb with the results, we swap back to `\` so that you get paths that match the system.  (Of course, this is a bit complicated, because UNC paths in windows have to remain with `/` chars.)

Such a change would have to come with tests, and I'd want to run those tests regularly to make sure they don't break.

ryanblock:
👋 
Re: https://github.com/isaacs/node-glob/issues/173#issuecomment-77702787 just got bitten by wrong path delimiters on Windows. But hey good news, GitHub Actions supports Windows. Would a GitHub Actions PR with Windows builds be helpful in working towards normalizing path delimiters per platform and closing this issue?

isaacs:
As of v9,

- paths are always returned in the results with their canonical platform-specific path separators
- glob patterns must use `/` for their path separation, unless `windowsPathsNoEscape` is used, in which case `/` is _only_ a path separator, and may not be used to escape glob characters.

PR #263, opened 2016-06-14
Document Windows returning improper path seperator

addisonj:


addisonj:
This isn't a fix for https://github.com/isaacs/node-glob/issues/173 but does at least document it...

Kapcash:
I was wondering if it was a normal behavior or a bug. Now I know this is expected to obtain slashes instead of backslashes on Windows!

PR #264, opened 2016-06-18
Fixed error in hasMagic function

akoutmos:
Fixed issue where empty pattern to hasMagic caused a TypeError.

Repro steps in node REPL:
```
> var glob = require('glob');
> glob.hasMagic('');
TypeError: Cannot read property 'length' of undefined
    at Function.glob.hasMagic (/........../node_modules/glob/glob.js:106:29)
    ...
```

### DIFF
--- a/glob.js
+++ b/glob.js
@@ -103,6 +103,9 @@ glob.hasMagic = function (pattern, options_) {
   if (set.length > 1)
     return true
 
+  if (set.length === 0)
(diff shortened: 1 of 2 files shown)


isaacs:
Already fixed, looks like?

akoutmos:
Just verified that this is indeed fixed. Thanks!

PR #271, opened 2016-07-07
Added appveyor support for node 6

amilajack:


jimmywarting:
Erm... have the balls to drop support for node less then 6 and go with nodes LTS release plan?
There are many using this module....

Issue #280, opened 2016-08-08
Add support to pass in fs interface


ghost:
I enjoy node-glob, but I need its use on a virtual file system which is separate from the node fs, but with the same interface. If we modify fs.realpath and add in a fs parameter in the options which falls back to the default native fs t should be quite easy

erikkemperman:
That would be exactly what I was looking for! Currently [`vinyl-fs`](https://github.com/gulpjs/vinyl-fs) is being updated to split off a module for use by other vinyl adapters, [`vinyl-prepare`](https://github.com/gulpjs/vinyl-prepare). With any luck it'll be in Gulp 4.

It would be great if implementers of custom vinyl adapters, such as purely in-memory ones, would not have to do their own globbing (or disallow glob patterns completely) -- but instead just provide the needed subset of `fs` calls as an option.

At this time, our call to glob can't be transplanted to `vinyl-prepare` because of its file system access, but I think it should be.

@isaacs I'd be willing to spend some time on a PR, unless of course you don't like the idea to begin with?

erikkemperman:
Come to think of it, maybe folks would like to use `graceful-fs`, that would also be easy with something like the above in place.

erikkemperman:
@Auxolust I've submitted a [PR](https://github.com/isaacs/node-glob/pull/285) that would address my use case... And yours, I believe?

Sorry about the repeated references above, I cleaned up the accompanying test a little bit and force-pushed to override the previous commit. But since it already had the issue number in there, this happened.

blond:
> Come to think of it, maybe folks would like to use graceful-fs, that would also be easy with something like the above in place.

@erikkemperman there are reasons to use `graceful-fs`: #299

millette:
I've got a [small patch](https://github.com/millette/node-glob/tree/opt-fs) too so I can use glob with dat and [hyperdrive](https://github.com/mafintosh/hyperdrive) and it seems to work for me so far. Basically, I can now use glob like this:
```
// dat.archive is a hyperdrive instance
glob('*', { cwd: '/', fs: dat.archive), (err, files) => {
  // and voilà
})
```

martinheidegger:
I was looking at this issue exactly because of how to use it with hyperdrive and plan to use the small patch by @millette . Total :+1: by me for having this. Would it be okay to publish a PR based on this prior work?

ajafff:
[...]

PR #285, opened 2016-09-05
Allow custom fs adapter to be passed as option (#280)

erikkemperman:
Maybe I should have allowed more time for negative feedback, but eager to get on with my planned refactor in `vinyl-fs` I figured I might as well take a stab at proposing something concrete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,8 @@ be immediately available on the `g.found` member.
 * `realpathCache` An optional object which is passed to `fs.realpath`
   to minimize unnecessary syscalls.  It is stored on the instantiated
   Glob object, and may be re-used.
+* `fs` By default `require('fs')` is used, but a custom filesystem
+  adapter may be provided as an option.
 
 ### Events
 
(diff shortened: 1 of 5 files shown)


erikkemperman:
Maybe point out the obvious: I haven't considered changes for the sync case yet. Also, it just occurred to me I've just assumed custom adapters would support `cache` argument to `realpath` but probably shouldn't.

erikkemperman:
I've added the synchronous counterpart, and tentatively simplified to pre-empt `realpath` calls entirely when a custom `fs` option is provided. Underestimated the implications of Node 6 slightly there.

Will leave this as it is for now, until you get a chance to look at this (or decide it's a bad idea), apologies for having submitted this a bit prematurely!

erikkemperman:
Rebased onto latest master.

erikkemperman:
Rebased again. Noticed failing test on Appveyor, have submitted a separate [PR](https://github.com/isaacs/node-glob/pull/303) which I believe addresses that.

@isaacs I appreciate you must be quite busy, but if possible I would be grateful for any quick feedback you might have here -- even if it's just that you don't like this idea... Thanks!

isaacs:
This will have to be added as an option to https://github.com/isaacs/path-scurry.  If so, it'll be trivial to pass it along from glob.

PR #293, opened 2016-09-16
Add `absolute` option for match event - closes #292

### LINKED ISSUE #292: Discussion: emitting absolute paths
In `glob-stream`, we do processing on each glob to resolve it to an absolute path before passing it to a globber.  This allows us to ensure we always have absolute paths emitted by this module.  However, I noticed that absolute paths are already being calculated (generated?) at https://github.com/isaacs/node-glob/blob/master/glob.js#L471

How would you feel about an added option, like `emitAbsolute`, to always emit the absolute path instead of the relative one?

phated:
@isaacs Implemented as per guidance on #292 - I added a note about the caveat that this is only applied to `'match'` events and not to the final results in the callback.  I am not familiar enough with the code to understand the scope of making the final results absolute.  Do you think it's a worth including in this feature?

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,9 @@ the filesystem.
   In the case of a symlink that cannot be resolved, the full absolute
   path to the matched entry is returned (though it will usually be a
   broken symlink)
+* `absolute` Set to true to always receive absolute paths for the
+  `'match'` event. Will not pass absolute paths to the callback
+  or returned from `glob.sync`.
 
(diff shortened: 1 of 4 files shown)


phated:
If there is anything else you'd like me to tackle as part of this, just let me know.

isaacs:
Published as 7.1.0.  Thanks!

phated:
Awesome!  Glad I could help find issues in `sync` too

Issue #245, opened 2016-01-26
files without permission entries are not found on Windows


schnittstabil:
Steps to reproduce:
1. create a file `deleteMe.txt`
2. `glob.sync('deleteMe.txt')` returns `[ 'deleteMe.txt' ]`
3. remove all permission entries (Windows Explorer)
4. `glob.sync('deleteMe.txt')` returns `[ ]` w/o error

(same with async…)

isaacs:
What do you get when you call `fs.statSync('deleteMe.txt')`?  Or run the glob with `glob.sync('deleteMe.txt', { stat: true })`?

schnittstabil:
1. `fs.statSync('deleteMe.txt')`

  ```
Error: EPERM: operation not permitted, stat 'C:\Dokumente und Einstellungen\Administrator\Eigene Dateien\deleteMe.txt'
    at Error (native)
    at Object.fs.statSync (fs.js:892:18)
    at Object.<anonymous> (C:\Dokumente und Einstellungen\Administrator\Eigene Dateien\test.js:6:31)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Function.Module.runMain (module.js:447:10)
    at startup (node.js:139:18)
    at node.js:999:3
```

2. `glob.sync('deleteMe.txt', { stat: true })` returns `[]`

3. But `fs.readdirSync('.')` returns `['deleteMe.txt', …]`

**Comment by isaacs** [comment]
Hm.  Ok, that seems kinda like a bug to me.  It should be handling `EPERM` from readdir calls as an indication that the thing is either not a directory or not traversable, but an `EPERM` on a file seems like it should be an error.

I'll have to review the code to be sure there isn't some method to this madness, but I agree that it seems like an oversight somewhere, and I'll try to figure out what bash does.  On Unix, there's no combination of permissions that will prevent you from being able to stat a file, so this is probably just a windows-ism that I didn't think to account for.

**Comment by mspace** [comment]
I came across the same issue: https://github.com/sindresorhus/del/issues/50. Also @schnittstabil  pointer that this is a glob issue.

**Comment by isaacs** [comment]
Anyone affected by this error, please try the following:

```
npm install isaacs/node-glob#fix-245
```

And let me know if the problem goes away.

isaacs:
[...]

PR #294, opened 2016-09-20
Add failing test for root + absolute option

phated:
Hey @isaacs - with the `absolute` option feature, I encountered a problem that I've created a failing test for.  I am still digging into a solution but I thought I'd get this test in front of you incase you have a better idea of what is happening.

Side note: I have a commented out test using `t.like` which was passing, not sure if I stumbled on a `node-tap` bug while working on this.

### DIFF
--- a/common.js
+++ b/common.js
@@ -103,7 +103,11 @@ function setopts (self, pattern, options) {
   if (process.platform === "win32")
     self.root = self.root.replace(/\\/g, "/")
 
-  self.cwdAbs = makeAbs(self, self.cwd)
+  // TODO: is an absolute `cwd` supposed to be resolved against `root`?
+  // e.g. { cwd: '/test', root: __dirname } === path.join(__dirname, '/test')
+  self.cwdAbs = isAbsolute(self.cwd) ? self.cwd : makeAbs(self, self.cwd)
+  if (process.platform === "win32")
+    self.cwdAbs = self.cwdAbs.replace(/\\/g, "/")
   self.nomount = !!options.nomount
 
   // disable comments and negation in Minimatch.
--- a/test/root.js
+++ b/test/root.js
@@ -81,3 +81,19 @@ t.test('combined with absolute option', function(t) {
     t.end()
   })
 })
+
+t.test('cwdAbs when root=a, absolute=true', function(t) {
+   var g = glob('/b*/**', { root: path.resolve('a'), absolute: true }, function (er, matches) {
+    t.ifError(er)
+    t.same(g.cwdAbs, process.cwd().replace(/\\/g, '/'))
+    t.end()
+  })
+})
+
+t.test('cwdAbs when root=a, absolute=true, cwd=__dirname', function(t) {
+   var g = glob('/b*/**', { root: path.resolve('a'), absolute: true, cwd: __dirname }, function (er, matches) {
+    t.ifError(er)
+    t.same(g.cwdAbs, __dirname.replace(/\\/g, '/'))
+    t.end()
+  })
+})
--- a/test/root.js
+++ b/test/root.js
@@ -52,3 +52,32 @@ t.test('root=a, cwd=a/b', function (t) {
     t.end()
   })
 })
+
+t.test('combined with absolute option', function(t) {
+  var g = glob('/b*/**', { root: path.resolve('a'), absolute: true }, function (er, matches) {
+    t.ifError(er)
+    /* For some reason this passes even though it compares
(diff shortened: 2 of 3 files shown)


phated:
Interesting that this passes on Windows.  That's extra confusing for me.

phated:
I think I have a partial commit (separate commit) but found another problem and added a failing test.

phated:
I think this actually might a symptom of a deeper problem with `makeAbs` because it seems `_emitMatch` is always receiving matches as absolute when the `root` option is set and caching results based on something like `/Users/phated/node-glob/test/fixtures/a/Users/phated/node-glob/test/fixtures/a/b` which seems to be wrong.

phated:
Rebased with possible fixes.  I have a TODO comment because I don't know the answer to:
> is an absolute `cwd` supposed to be resolved against `root`?
e.g. { cwd: '/test', root: __dirname } === path.join(__dirname, '/test')

phated:
Hmm, broke something else on Windows.

phated:
Yay, passing tests!  I'm not sure if everything I did to get these tests passing is kosher or even that the tests are written correctly, but it seems to be doing the right thing in the few narrow cases I came up with.

Can you think of a better way to write the `cacheCheck` tests to verify the paths being cached actually exist and isn't suffering from this joining of 2 absolute paths?

test/root.js
```diff
@@ -52,3 +52,48 @@ t.test('root=a, cwd=a/b', function (t) {
...
+        '/Users/phated/node-glob/test/fixtures/a/Users/phated/node-glob/test/fixtures/a/bc/e/f' ]
+      to
+      [ '/Users/phated/node-glob/test/fixtures/a/b',
+        '/Users/phated/node-glob/test/fixtures/a/b/c',
+        '/Users/phated/node-glob/test/fixtures/a/b/c/d',
+        '/Users/phated/node-glob/test/fixtures/a/bc',
+        '/Users/phated/node-glob/test/fixtures/a/bc/e',
+        '/Users/phated/node-glob/test/fixtures/a/bc/e/f' ]
+    */
+    // t.like(matches, [ '/b', '/b/c', '/b/c/d', '/bc', '/bc/e', '/bc/e/f' ].map(function (m) {
```
isaacs: Ha!  Yes, this is because `t.like` does substring matching.  So, `t.like('hamburger', 'amburg')` passes.  Calling `t.like(array, otherArray)` does a `t.like` on the elements, so `t.like(['hamburger'], ['amburg'])` would also pass.  `tap` is working as designed, but I'm using it wrong here.

test/root.js
```diff
@@ -52,3 +52,48 @@ t.test('root=a, cwd=a/b', function (t) {
...
+    //   return path.join(path.resolve('a'), m).replace(/\\/g, '/')
+    // }))
+    t.same(matches, [ '/b', '/b/c', '/b/c/d', '/bc', '/bc/e', '/bc/e/f' ].map(function (m) {
```
isaacs: `t.same` is the right thing to use here.

common.js
```diff
@@ -103,7 +103,11 @@ function setopts (self, pattern, options) {
   if (process.platform === "win32")
     self.root = self.root.replace(/\\/g, "/")
 
-  self.cwdAbs = makeAbs(self, self.cwd)
+  // TODO: is an absolute `cwd` supposed to be resolved against `root`?
+  // e.g. { cwd: '/test', root: __dirname } === path.join(__dirname, '/test')
+  self.cwdAbs = isAbsolute(self.cwd) ? self.cwd : makeAbs(self, self.cwd)
```
isaacs: This seems like a reasonable fix.  We do need `makeAbs` to mount onto the root, if one is specified, so that `/a/*` maps onto `{root}/a/b`.  But using cwd and root together is just... kinda weird.

test/root.js
phated: alright, I'll update the tests in this file to use `t.same`

common.js
phated: Hmm, how should I go about resolving this then?  If `makeAbs` is used, it appends an absolute path to an absolute root resulting in something like `/Users/phated/node-glob/test/Users/phated/node-glob/test/fixtures`

phated:
@isaacs sorry to ping again but I'd like to get this resolved in some way so I can land the usage in `glob-stream`

isaacs:
Thanks! this'll go out in the next release.  I also wanna fix the thing with the stat errors on Windows, in #245.

PR #300, opened 2016-10-09
configuration to enable glob to support bower

jamesdbloom:
adding configuration for bower to enable glob to be installed using bower


isaacs:
I'm not interested in maintaining version, description, and dependency information in more than one place.

You are welcome to float this patch on your own branch, and use that with bower.

PR #303, opened 2016-10-26
Fix win32 test failure

erikkemperman:
The test in `eperm-stat.js` fails on Appveyor because the fixtures don't include symlinks. This PR remedies that.

### DIFF
--- a/test/eperm-stat.js
+++ b/test/eperm-stat.js
@@ -87,6 +87,11 @@ t.test('globstar with error in root', function (t) {
     'a/x',
     'a/z'
   ]
+  if (process.platform === 'win32') {
+    expect = expect.filter(function(path) {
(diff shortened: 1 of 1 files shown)


raygervais:
With this fix, I believe many of the PRs of recent will pass the tests now. Hopefully you get merged sooner than later.

erikkemperman:
@raygervais All due respect to the maintainer, but it looks like he doesn't get around to handling issues/PRs very often. Kind of troubling, if you think about the very many packages depending on this.

jdalton:
@isaacs As part of my follow-up from https://github.com/npm/node-semver/pull/244 this PR would allow unskipping in CitGM for Windows _(restores passing appveyor too)_.

PR #304, opened 2016-10-29
Drop `inherits` dependency in favor of `util.inherits` in Node itself.

isiahmeadows:
Not like this supports any other runtimes, anyways...


isiahmeadows:
Failing test appears unrelated.

isiahmeadows:
@isaacs Ping?

PR #308, opened 2016-11-11
Removed unused variables

jackmoore:


isaacs:
No longer an issue in v9.

PR #310, opened 2016-11-21
:bug: Always emit matches asynchronously

gzzhanghao:
Glob may emit match events synchronously when cache are given. And it leads to a race condition, eg:

```javascript
let glob = new Glob(globStr, { cache, symlinks, statCache })

// Wait for glob ends...

glob = new Glob(globStr, { cache, symlinks, statCache }) // <- Emit `match` event immediately
glob.on('match', console.log) // <- Receive nothing
```

### DIFF
--- a/glob.js
+++ b/glob.js
@@ -168,10 +168,12 @@ function Glob (pattern, options, cb) {
     return done()
 
   var sync = true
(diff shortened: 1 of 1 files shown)


isaacs:
No longer an issue in v9.

PR #311, opened 2016-12-15
Clarify optional parameters

mikelambert:
Clarify that optional means it can be omitted from the argument list, by marking the comma optional as well.


isaacs:
No longer relevant in v9.

Issue #323, opened 2017-03-02
'end' is emitted on pause, not on completion


jmwilkinson:
As the title says, if you listen for the end event, it will be emitted if the Glob object is paused. It will be emitted only once, on the first pause. It will not be emitted when the Glob object completes.

bitstrider:
This seems to happen sometimes with certain configurations and directory structures.  I'm currently working on a PR with a test case for this bug, will post soon.

bitstrider:
#328

isaacs:
No longer relevant in v9

PR #328, opened 2017-03-31
Test to show if 'end' event is emitted properly

json2d:
- addressing #323 
- adds two new test patterns: a/\*, a/\*/\*\*
- includes two variants of the test, using deep and shallow result
comparison
- the deep variant shows when 'end' misfires


isaacs:
No longer relevant in v9.

PR #330, opened 2017-04-06
Fixes #329

### LINKED ISSUE #329: test/eperm-stat.js should be ignored on win32
`test/eperm-stat.js` has symlinks in its result, so it should be ignored, like in other tests, on Windows.

bitstrider:
simple addition to make tests pass on win32


bitstrider:
Duplicate of #303, closing

PR #336, opened 2017-05-01
Add option to cache lstat results instead of stat

erikkemperman:
I'm preparing a PR to vinyl-fs to signifcantly reduce the number of fs operations by leveraging node-glob's caches. However there's a `resolveSymlinks` option which users set to false to obtain stats of the link, rather than the target, for symlink matches.

Currently this has us do three stats (an lstat and stat in node-glob, and then another lstat in vinyl-fs). This PR, or something like it, would allow me to reduce that to just one (lstat).

PS I've included a commit from another [PR](https://github.com/isaacs/node-glob/pull/303) to make Appveyor pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,8 @@ the filesystem.
   to set this, you may pass the statCache from one glob() call to the
   options object of another, if you know that the filesystem will not
   change between calls.  (See "Race Conditions" below.)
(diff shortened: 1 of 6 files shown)


isaacs:
No longer relevant in v9.

PR #345, opened 2017-07-13
Add support for UNC paths

bdukes:
This adds *nix support for the test in @staticshock's PR #146 (test fails on *nix before the implementation and passes on *nix after the implementation)

The test operates on *nix by mocking out `process.platform` (set to `'win32'`), `minimatch.Minimatch` (to fix differences between slashes) and `path.resolve` (to translate the UNC path to the local mapped path).

This also removes the `nocase` test from #146 because there was not a working implementation, though I hope to be able to add an implementation and test there, too.


isaacs:
Fixed in v9.

PR #346, opened 2017-07-14
Match absolute paths on Windows

### LINKED ISSUE #290: C:\\* returns empty list
hi,
if you pass `C:\\*` or `C:/*` its returning nothing. Is there are trick to get the folder/file list on a root disc ? 
thanks,
g

bdukes:
This fixes a bug on Windows where `root` is mishandled and so it is impossible to match absolute paths.  This PR depends on #345.

Fixes #290


isaacs:
Fixed in v9.

PR #349, opened 2017-09-01
correct spelling mistake

EdwardBetts:


test/empty-set.js
```diff
@@ -7,7 +7,7 @@ var patterns = [
   '# comment',
   ' ',
   '\n',
-  'just doesnt happen to match anything so this is a control'
+  'just doesn't happen to match anything so this is a control'
```
Nuuou: Looks like the single quote here needs to be escaped, causing a syntax error otherwise.

EdwardBetts: Sorry about that. I will close this request.

PR #352, opened 2017-10-18
Fixes #351, doesn't mark path as 'FILE' unless it exists

### LINKED ISSUE #351: Glob.sync results include files that don't exist
I've added a reproduce repository [here](https://github.com/mitchhentges/node-glob-returning-file-that-doesnt-exist).

Sometimes, `glob.sync()` will return paths to files that don't actually exist. For example, with the following tree:

```
$ tree -a
.
├── index.js
└── node_modules
    └── @namespace
        └── src-package-a
            └── node_modules
                └── lib-b
                    └── node_modules
                        └── lib-c
                            └── bin
                                └── file.js
```

The glob `'node_modules/**/node_modules/**/node_modules/**/node_modules'` will return  `node_modules/@namespace/src-package-a/node_modules/lib-b/node_modules/lib-c/bin/file.js/node_modules`, which doesn't actually exist.

Node `v6.11.3`, MacOS `10.12.6`

-----

This is because upon a `readdir` error happening, `glob.sync` assumes that a `ENOTDIR` error means that a path points to a file. However, in the case where a path _includes_ a file (e.g. `/etc/passwd/some_directory`, where `passwd` is the file in the middle of the path), `fs.readdir()` will throw `ENOTDIR`, not `ENOENT`.
The solution here is: when catching an `fs.readdir()` error, if it is of type `ENOTDIR`, verify that the result is indeed a real file, and not an invalid path with a file in the middle.

mitchhentges:


mitchhentges:
I didn't add an additional test for this because:
* `npm run test-regen` is failing locally
* Tests are failing in CI

So, it sounds like there's existing test issues

mitchhentges:
This repository hasn't been updated in almost a year, and there's a lot of PRs in purgatory 😢 
However, if you're reviewing this PR as a maintainer, let me know! I just don't want to be spending time polishing a PR if it's for a dead project :smile:

PR #364, opened 2018-03-17
Fix ENOTDIR error in sync.js when reading directories

saurini:
This was causing an issue for me on Ubuntu 16.04 with NVM and Node 8.10 when running `yo`.


isaacs:
No longer relevant in v9.

PR #381, opened 2018-11-03
add CLI

fabiosantoscode:
This adds a CLI interface. Handy for finding files in recursive directories in a bash shell

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -1,0 +1,9 @@
+#!/usr/bin/env node
+
+var glob = require('.')
+
+process.argv.slice(2).forEach(function (arg) {
+  glob.sync(arg).forEach(function (line) {
+    console.log(line)
+  })
+})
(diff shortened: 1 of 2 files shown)


markstos:
I recommend closing this since `glob-cli2` already exists and uses this module:

https://github.com/a0viedo/glob-cli2/blob/master/index.js

markstos:
@fabiosantoscode If you want to help with the `glob` CLI, you could implement this feature request to return propertly quoted strings when requested:

https://github.com/a0viedo/glob-cli2/issues/2

This is helpful in cases where file or path names have spaces in them.

fabiosantoscode:
Got it. Not interested in contributing to the glob cli though.

gkop:
Just wanted to let anybody else that visits this issue and is considering trying glob-cli2, that unfortunately it is extraordinary limited as [it does not permit passing options](https://github.com/a0viedo/glob-cli2/blob/bf160f1aafd509f08017b12d2364bdbaf6f92671/index.js#L7) to glob.

Issue #380, opened 2018-10-16
Failing CI due to old Node.js versions


DanielRuf:
![bildschirmfoto 2018-10-16 um 10 04 10](https://user-images.githubusercontent.com/827205/47001488-dd834780-d12a-11e8-9018-b63034edbb17.png)

![bildschirmfoto 2018-10-16 um 10 04 40](https://user-images.githubusercontent.com/827205/47001515-ed9b2700-d12a-11e8-9993-cf2b1305eacc.png)

![bildschirmfoto 2018-10-16 um 10 04 45](https://user-images.githubusercontent.com/827205/47001516-ee33bd80-d12a-11e8-9b89-3a43ec0547e2.png)

DanielRuf:
It seems this was silently fixed a while ago now.

Issue #376, opened 2018-09-05
Drop support for node 0.10, 0.12 and maybe more?


jimmywarting:
I wish you dropped support for eol versions and only targeting LTS version so that you can for example drop inherits and instead use `class Gulp extends EE {}`

as a new Major version update where you also drop all deprecated api (old api surface) code

all test in travis fails due to `update-notifier` using `let`

the-spyke:
And replace callbacks with Promises? 😅

PR #385, opened 2018-12-23
require Node 6+ and remove `path-is-absolute` dependency

Krinkle:
Node 4 and earlier are EOL, and the polyfill package has been
deprecated and its repository archived.

Fixes https://github.com/isaacs/node-glob/issues/376.
Fixes https://github.com/isaacs/node-glob/issues/380.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - '0.10'
-  - '0.12'
-  - '4'
-  - '5'
   - '6'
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -1,8 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - '0.10'
-  - '0.12'
-  - '4'
-  - '5'
   - '6'
```
jimmywarting: 6 can be dropped now...

PR #397, opened 2019-06-06
Fix documentation for synchronous functionality

YodasWs:
Because the user cannot merely remove the callback function for the blog to be done synchronously. What *actually* makes it run synchronously is setting `sync: true` in `options`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,8 +127,7 @@ Perform an asynchronous glob search.
 
   :warning: Must include property `sync: true` to run synchronously.
 
-* return: `{Object}`
-  * `matches` `{Array<String>}` Array of matched filenames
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,11 +120,14 @@ options.
 
 Perform an asynchronous glob search.
 
-## glob.sync(pattern, [options])
+## glob.sync(pattern, options)
```
hg-pyun: `[ ]` means optionality. I think it need.

YodasWs: But in order to run this *asynchronously*, you need to include the `options` object.

isaacs:
No longer relevant in v9.

PR #398, opened 2019-06-17
Remove .gif that is not used in README any more

zakhenry:


isaacs:
End of an era :)

zakhenry:
Hah I have to admit I couldn't remember the name of this package but I did remember "oh my glob" which did get me search results when I google `oh my glob node` 😁

PR #403, opened 2019-10-12
Use path.isAbsolute

ip1981:
The path-is-absolute package is no longer relevant as Node.js 0.12 is
unmaintained.


ip1981:
Meh, every year => https://github.com/isaacs/node-glob/pull/385 :)

PR #404, opened 2019-10-23
Remove unused variables

abetomo:
I found some unused variables and removed them.


isaacs:
No longer an issue in v9.

PR #411, opened 2020-03-21
chore: Remove nested FUNDING.yml

nschonni:
Looks like a second FUNDING was adding in an additional `.github` folder by accident


isaacs:
Fixed some time ago.

PR #412, opened 2020-03-25
build: Use GitHub Actions for multi-os CI

nschonni:
Left in 8, even though it's EOL and added 13.
Also expanded to multi-os build, so the AppVeyor hook can be turned off

Sample run can be found here https://github.com/nschonni/node-glob/actions/runs/62701471


isaacs:
Done in main

Issue #298, opened 2016-10-07
`options.strict` default value


schnittstabil:
As far as I can see, the default value of `strict` is `false`:
[`self.strict = options.strict !== false`](https://github.com/isaacs/node-glob/blob/cea11fea473f18c565a515b0c61c8f194fc23aaa/common.js#L67).

Which sounds sensible to me, btw. However, in the [readme](https://github.com/isaacs/node-glob/tree/cea11fea473f18c565a515b0c61c8f194fc23aaa#options):

> All options are false by default, unless otherwise noted.
> …
> - `strict` When an unusual error is encountered when attempting to read a directory, the process will just continue on in search of other matches. Set the `strict` option to raise an error in these cases.

isaacs:
No longer an issue on v9

PR #413, opened 2020-04-18
Update strict documentation to clarify default value

skovhus:
Currently, strict defaults to true: https://github.com/isaacs/node-glob/blob/b2aa29bc066de1593d78864158e6a1adafd06e3c/common.js#L67

As `undefined !== false` is true. This makes sense, but it isn't clear in the documentation where "All options are false by default, unless otherwise noted."

Fixes https://github.com/isaacs/node-glob/issues/298


isaacs:
No longer relevant in v9.

skovhus:
@isaacs thanks. And really great work on version 9! 🙏

PR #430, opened 2020-11-18
Add fs option to allow passing virtual filesystem

### LINKED ISSUE #429: Virtual file system option?
It'd be great if we could pass in a virtual filesystem as an option. There are several compatible with the `fs` API so it shouldn't require many changes:

- https://github.com/streamich/memfs
- https://github.com/jvilk/BrowserFS
- https://github.com/filerjs/filer (no sync API)
- https://github.com/isomorphic-git/lightning-fs (no sync API)

How it'd work:

```js
let glob = require('glob')
let memfs = require('memfs')

let volume = memfs.Volume.fromJSON({
  '/foo/bar/abc.txt': 'abc',
  '/foo/bar/def.log': 'def',
  '/foo/bar/xyz.txt': 'xyz',
})

glob.sync('/foo/bar/*.txt', { fs: volume }) // returns ['/foo/bar/abc.txt', '/foo/bar/xyz.txt']
```

This would allow usage of `glob` in lots more places without adding any real maintenance burden or requiring users to monkey-patch. All of the hard work would be done by one of the above libraries, and glob could be used in the browser, with dropbox, S3, and many others.

It'd also be useful for speeding up the testing of libraries that rely on `glob` - instead of creating fixtures that touch the real filesystem, the same operations could be done with in-memory objects using `memfs` (although monkey-patching is less risky in test code).

Of course, to make sure it doesn't add overhead to maintenance of this library, it should come with the disclaimer that `glob` doesn't take responsibility for incomplete/imperfect `fs`-emulation. If users have issues with symlinks/missing functionality, they should raise an issue with the library being used to get a virtual fs rather than glob.

mmkal:
Fixes #429 

___

This enables passing in a virtual filesystem compatible with the `fs` module. There are several matching the `fs` API:

- https://github.com/streamich/memfs
- https://github.com/jvilk/BrowserFS
- https://github.com/filerjs/filer (no sync API)
- https://github.com/isomorphic-git/lightning-fs (no sync API)

Example:

```js
let glob = require('glob')
let memfs = require('memfs')

let volume = memfs.Volume.fromJSON({
  '/foo/bar/abc.txt': 'abc',
  '/foo/bar/def.log': 'def',
  '/foo/bar/xyz.txt': 'xyz',
})

glob.sync('/foo/bar/*.txt', { fs: volume }) // returns ['/foo/bar/abc.txt', '/foo/bar/xyz.txt']
```

Allows usage of `glob` in lots more places without requiring users to monkey-patch. With this change, glob could be used in the browser, with dropbox, S3, purely in memory like above (useful for testing) and many others.

___

After writing that issue up, I had a look at the code and the change required is very small so I thought I'd just open a PR.


isaacs:
Thanks, landed and published as 7.2.0

PR #437, opened 2021-03-08
Provide a custom fs module option

Kleywalker:
To provide a custom `fs` module like the in-memory file-system [memfs](https://github.com/streamich/memfs) often used for testing.
  Example:
  ```js
  var glob = require("glob");
  var vol = require("memfs").vol;
  const json = {
    "./README.md": "1",
    "./src/index.js": "2",
    "./node_modules/debug/index.js": "3",
  };
  vol.fromJSON(json, "/app");
  var opt = {
    cwd: "/app",
    fs: vol,
  };
  glob("**/*.js", opt, function (er, files) {
    if (er) {
      console.log(er);
    }
    console.log(files);
  });
  ```

### DIFF
--- a/glob.js
+++ b/glob.js
@@ -123,6 +123,10 @@ function Glob (pattern, options, cb) {
     options = null
   }
 
+  if (options && options.fs) {
+    fs = options.fs
+  }
+
   if (options && options.sync) {
     if (cb)
       throw new TypeError('callback provided to sync glob')
--- a/README.md
+++ b/README.md
@@ -276,6 +276,31 @@ the filesystem.
 * `absolute` Set to true to always receive absolute paths for matched
   files.  Unlike `realpath`, this also affects the values returned in
   the `match` event.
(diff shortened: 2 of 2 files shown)


glob.js
```diff
@@ -123,6 +123,10 @@ function Glob (pattern, options, cb) {
     options = null
   }
 
+  if (options && options.fs) {
+    fs = options.fs
```
mmkal, line 127: this will override it permanently.

```js
glob('*.txt', { cwd: '/some/directory', fs: vol }, function (e, f) {...})

glob('*.txt', { cwd: '/some/directory' }, function (e, f) {...})
```

👆 the second call here will use the virtual file system, even though it should use the real one.

mmkal (changes requested):
This is a duplicate of a PR I opened a few months ago: https://github.com/isaacs/node-glob/pull/430

The other one has tests, has included an implementation for the `glob.sync` API too, and avoids the bug mentioned inline about reassigning the `fs` variable beyond the scope of a single call. So I think this one should be closed.

PR #451, opened 2022-01-05
[Refactor] use more explicit `assert.ok`

ljharb:
This is a tiny refactor, hopefully it's acceptable.

(We'll all be much happier if you choose not to ask why i need this change :-) )

### DIFF
--- a/sync.js
+++ b/sync.js
@@ -48,7 +48,7 @@ function GlobSync (pattern, options) {
 }
 
 GlobSync.prototype._finish = function () {
-  assert(this instanceof GlobSync)
+  assert.ok(this instanceof GlobSync)
   if (this.realpath) {
     var self = this
     this.matches.forEach(function (matchset, index) {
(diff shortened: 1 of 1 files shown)


nicolas377:
@isaacs does this need documentation since it's such a small change?

ljharb:
I'd assume not; virtually nobody but me's going to notice.

ljharb:
@nicolas377 see the second sentence in the OP.

In node 0.6+, this will behave identically. Absolutely nothing needs to be documented that affect an older version of node than that :-)

nicolas377 (approved):
@isaacs LGTM! All the functionality this changes is out of scope of this package's supported node versions.

isaacs:
> (We'll all be much happier if you choose not to ask why i need this change :-) )

Oh god, I think I know what's going on here lol 😂 

Do you need this backported to v7?  v8 is node 12+ already anyway.

ljharb:
@isaacs ideally yes, but i understand if you're unwilling. either way this drastically reduces the number of n_m patches i need to float :-)

isaacs:
Published on 7.2.2 and 8.0.3

PR #466, opened 2022-04-11
fix: bump version packages due to security issue PRISMA-2022-0039

opravil-jan:


isaacs:
No longer relevant in v9.

PR #473, opened 2022-05-14
fix: disable allowWindowsEscape on minimatch

### LINKED ISSUE #471: Coordinated fix on Feb 12 broke Windows path handling between node-glob and minimatch
When using `allowWindowsEscape=true` in minimatch options, glob is forcing minimatch to treat the path as though it contained _escapes_ rather than Windows path separators (backslashes). Notice the negation in the following statement, which would entirely fix the issue if it were allowed to run.

```js
  if (!options.allowWindowsEscape && path.sep !== '/') {
    pattern = pattern.split(path.sep).join('/')
  }
```

Therefore, `glob@7.2.2` combined with `minimatch@^3.1.1` _cannot_ locate files using patterns like:

    E:\\Src\\repo\\packages\\test\\.storybook\\main{'.js','.jsx'}

Unfortunately, this is causing the latest version of Storybook to fail to locate configuration files in its default state in Windows. Error: No configuration files have been found in your configDir.

Please coordinate another fix.

nicolas377:
This fixes #471 by a partial revert of 73feafd17f1c04633349652605e12fbbaef9b3cf.
See my comment on the issue https://github.com/isaacs/node-glob/issues/471#issuecomment-1126739067 for more details.

I intend for this to be released as a patch to only v7.

### DIFF
--- a/common.js
+++ b/common.js
@@ -110,8 +110,6 @@ function setopts (self, pattern, options) {
   // Note that they are not supported in Glob itself anyway.
   options.nonegate = true
   options.nocomment = true
-  // always treat \ in patterns as escapes, not path separators
-  options.allowWindowsEscape = true
 
   self.minimatch = new Minimatch(pattern, options)
   self.options = self.minimatch.options


nicolas377:
Fix landed in c663c7d21ffe64ddb5576674735c65a8551f927a.

isaacs:
Ah, sorry, didn't see this before fixing it myself.  Thanks, though!

nicolas377:
All good!

Issue #468, opened 2022-04-26
feat: coerce all paths to forward slashes as a config option


nicolas377:
With the recent issues in glob v8 (see the conversation in #467), I thought it would be a good idea to add a config option that would coerce the inputted glob path to use forward slashes instead of just forcing users to convert their paths to forward slashes.

Some use cases allow users to predictably use forward slashes (e.x. using relative paths in glob calls), but most don't allow that.

I'm not sure of a name for the config option, I'm open to suggestions. Once I get approval from the owners, I'll go ahead and open a PR.

nicolas377:
Just clarifying, the config option would be disabled by default.

isaacs:
How about this: `glob('c:\\some\\path\\*.js', {windowsPaths: true})`
or: `glob('c:\\some\\path\\*.js', {pathSeparator: '\\'})`

With the caveat that if you set `windowsPaths: true` (or `pathSeparator:'\\'`) then there is no way to escape glob characters, meaning it is _impossible_ to ever match a file literally named `*` or `?`.

nicolas377:
I like `pathSeparator`, especially since it can be easily set to [path.sep](https://nodejs.org/api/path.html#pathsep).

Most of my use cases for glob involve resolving paths to an absolute path, and then appending to that string. I'm not sure if my use case is common, but it seems like it would be common, and I think that `pathSeparator` would help a lot with that use case.

ath0mas:
I like `pathSeparator` with `path.sep` too,
but it may feel "strange" and be misleading for users to have to set it explicitly while the most common or expected implicit default value?

> then there is no way to escape glob characters, meaning it is impossible to ever match a file literally named `*` or `?`

or "better" be an option to explicitly disable this possibility, `disableGlobCharMatch` or so, to allow to match "like before" without any replace or option about the "path separator" thing?

isaacs:
So you can explicitly opt in to using \ as your path separator (and not have escaping), or maybe we make both the path separator _and_ the escape character configurable? But "just use the implicit default path separator, and also be a bash style glob" isn't logically coherent.

nicolas377:
The issue that I was describing was when dealing with paths run through either path.join or path.resolve. Most other cases still work, and don't need changes.

nicolas377:
> couldn't you just replace every \\ by / whenever the platform is Windows

That's actually pretty close to the implementation I was thinking of. My idea of the implementation of `pathSeparator` would go something like this:

[...]

PR #475, opened 2022-05-15
feat: add pathsep option

nicolas377:
This is the first part of the implementation of #468. This PR implements the pathsep option, which is intended to allow users to specify the path separator used in the glob pattern.

### DIFF
--- a/common.js
+++ b/common.js
@@ -83,6 +83,11 @@ function setopts (self, pattern, options) {
   self.statCache = options.statCache || Object.create(null)
   self.symlinks = options.symlinks || Object.create(null)
 
+  // avoiding empty strings
+  if (!!options.pathsep) {
+    pattern = pattern.split(options.pathsep).join("/")
+  }
+
   setupIgnores(self, options)
 
   self.changedCwd = false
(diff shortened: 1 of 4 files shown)


nicolas377:
Tests are passing on [the fork](https://github.com/nicolas377/node-glob/actions/runs/2327795915).

common.js
```diff
@@ -83,6 +83,11 @@ function setopts (self, pattern, options) {
   self.statCache = options.statCache || Object.create(null)
   self.symlinks = options.symlinks || Object.create(null)
 
+  // avoiding empty strings
+  if (!!options.pathsep) {
+    pattern = pattern.split(options.pathsep).join("/")
```
nicolas377, line 88: While using String.prototype.replace with a plain RegExp with the `g` flag (String.prototype.replaceAll isn't supported on node v12 and v14) is significantly faster than splitting and joining strings, creating a RegExp from dynamic content poses both security risks and can throw incorrect errors when a single backslash is inputted. 
Escaping potentially dangerous characters will put a large load on one point security-wise, and is less performant than simply splitting and joining strings.

isaacs:
No longer relevant, landed on `windowsPathsNoEscape` option.

PR #476, opened 2022-05-15
feat: add supportWindowsPaths option

### LINKED ISSUE #468: feat: coerce all paths to forward slashes as a config option

nicolas377:
This is the second and last part of the implementation of #468. The PR implements the supportWindowsPaths option, which is intended to be a flip-on switch for automatic backslash coercion.

closes #468

### DIFF
--- a/common.js
+++ b/common.js
@@ -68,6 +68,11 @@ function setopts (self, pattern, options) {
   self.nodir = !!options.nodir
   if (self.nodir)
     self.mark = true
+  self.supportWindowsPaths = !!options.supportWindowsPaths
+  if (self.supportWindowsPaths) {
+    options.pathSep = undefined
+    pattern = pattern.replace(/\\/g, "/")
+  }
   self.sync = !!options.sync
   self.nounique = !!options.nounique
   self.nonull = !!options.nonull
--- a/README.md
+++ b/README.md
@@ -240,6 +240,9 @@ the filesystem.
   to set this, you may pass the statCache from one glob() call to the
   options object of another, if you know that the filesystem will not
   change between calls.  (See "Race Conditions" below.)
+* `supportWindowsPaths` Supports both `/` and `\` as path separators.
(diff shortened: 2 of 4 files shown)


common.js
```diff
@@ -68,6 +68,11 @@ function setopts (self, pattern, options) {
   self.nodir = !!options.nodir
   if (self.nodir)
     self.mark = true
+  self.supportWindowsPaths = !!options.supportWindowsPaths
+  if (self.supportWindowsPaths) {
+    options.pathSep = undefined
+    pattern = pattern.replace(/\\/g, "/")
```
nicolas377, line 74: Cool things I found out, `/\\/g` is **105% faster** than `/\\|\//g`.

isaacs: So instead of doing 1000/ms you can do 2000?  Not trying to be too snarky, but it's a one-time action at initialization, and wildly fast, even in a benchmark with 2000 path portions and a mix of `\` and `/`.  Even if it was 10,000x faster, even if it was somehow able to magically do it in 0 CPU cycles, it's still only saving you 0.0001% of the overall time.

isaacs: Of course, if it was actually possible to do in 0 CPU cycles, we'd be better off spending our time figuring how how to express all compute problems in terms of slash replacement, and win the turing prize for solving P=NP lol

nicolas377: I know that it's a very minor improvement in the grand scheme of things, it just seemed pretty cool to me.

isaacs:
This landed as `windowsPathsNoEscape` some time ago.

PR #489, opened 2023-01-16
V9

isaacs:
As part of my mission to rewrite all my modules as TypeScript modules with hybrid exports, and update the entire world apparently, v9 is coming.

It's pretty much working, but the tests still need a lot of cleanup.

Summary of changes [in the changelog](https://github.com/isaacs/node-glob/blob/v9/changelog.md#90).


isaacs:
100% test coverage all passing!

The only thing left to do is refactor some of the initial path management out of the walker class.  Right now, it's a bit too smart.  All of the "setStart()" and absolute/windows path detection belongs in Glob, not GlobWalker.

isaacs:
Mostly got performance up to where I want it with some aggressive caching of readdir calls and a few other minor fixes, but found an interesting pathological cases.  When a globstar is followed by multiple magic patterns, especially if those include another glob star, the performance falls off a cliff.

So, in a big directory tree, a pattern like `**/*/*.txt` takes about twice as long as it should, and a pattern like `**/*/**/*.txt` takes 3x as long as it should.

Need to rethink the approach for handling `**`, I think.

isaacs:
It is currently spending too much time in path resolution. I have an idea for a way around it, though it does involve writing Yet Another dir walker type thing, as none of the ones I found on npm give me quite what I want. But, I think that could be useful if I ever need to revisit modules like ignore-file that do directory traversals, so it would be worth being it's own thing.

meyfa:
Thanks for working on this, and for your recent work on rimraf and minimatch, too! These dependencies are in literally all of my projects, often multiple times. It's not at all a given that you'd spend your time on this, but you do, and it's improving the dev experience for many thousands of people. Thank you!

isaacs:
This exists now: http://npm.im/pointer-set

Now on to the thing that uses it.

isaacs:
Ok, so, TURNS OUT, using uint32 slab pointers in this way *can be* faster, *but* only in cases where gc is a major performance killer, ie where objects are being deleted quickly, like in a cache that's pushing out old items.  That's why it makes lru cache implementations so unreasonably fast.  But in a case where I'm walking a bunch of paths _and keeping the information around for the duration of the walk_, the work of allocating and accessing and dereferencing uint32 pointers into an expanding data store of that sort is actually much _worse_ than just using consistently-shaped objects, since they're not going to be garbage collected until after the call is completed.

And in fact, if you re-use the data store (so a reference to it is maintained), then it's never garbage collected _ever_ so the object-based data store (which, also, is like, _way_ simpler and less awkward to work with) is significantly faster.

All this is to say, the exploration of pointer-set was fun to write and might be useful for some application someday, but it's not useful _for this_, which means it was... well, not pointless, but academic rather than practical.

On to the next approach, I guess, the (not pointer-based) path walker.  The fundamental algorithm is still worth pursuing, just don't need to be doing as much pointer arithmetic and bit-shifting, which frankly, is kind of a relief in a way.

isaacs:
Since I know some folks are watching this, maybe some of y'all could weigh in:

1. Dirent data: Would it be worthwhile to have a `{ withFileTypes: true }` option, and get a Dirent-type thing returned instead of path strings?  I find a lot of applications where I'm globbing for paths, I immediately turn around and stat them to call `isDirectory()` or something on them, so it might be handy to have that already done, since the glob walk has to gather that info on its way anyhow.

2. Bash-style weird paths.  Let's say you have a folder structure like:
    
    ```
    a
    +-- b
    |   +-- c
    +-- x
    ```

    If you do `echo a/*/../b/c` then shells will output `a/b/../b/c a/x/../b/c`, even though these two entries are actually the same thing.  Would it be more useful to just get a single `a/b/c` back?  That'd actually be much easier to do.

    Same if you have dots or multiple slashes in the pattern.  So if you did `a///.///*/c`, that'll return `a///./b/c`.  Just returning the joined path would be less work, because it would mean that I can just walk up and down the tree that PathScurry creates, which cuts out the need to track the literal pattern that got resolved at each step.

Bonus, this is going to handle UNC paths and windows drives, and in windows mode, will return canonical windows-style paths, like `c:\\x\\y\\z` instead of `c:/x/y/z`. Tbh, as a long time unix user, the second looks way less weird to me, but I know feels off-putting to Windows users, and PathScurry makes that the default behavior.

SimenB:
1. `{ withFileTypes: true }` would be very useful! 👍 
2. The second seems nicer and more useful in practice with your suggestion - how much to diverge from shell globbing and how that might surprise people I couldn't say (IIRC I've never used either syntax). Maybe a `normalize` (or something) option which defaults to `true`?

isaacs:
> The second seems nicer and more useful in practice with your suggestion - how much to diverge from shell globbing and how that might surprise people I couldn't say (IIRC I've never used either syntax). Maybe a normalize (or something) option which defaults to true?

I could see it being a bit weird if you get absolute paths instead of relative paths.  For example, if you had a module that globs for files to include in a `package.json` or other project-specific config file, then I *definitely* don't want to end up hard-coding `/Users/isaacs/projects/isaacs/blah/dist/foo.js` somewhere, and having to do an extra step to relativize the paths is annoying.

Tracking the "pattern expanded" path is a bit odd, tbh, and results in getting duplicate entries, which isn't a huge deal in a shell context, but can be annoying in other places.  (Eg, getting both `a/b/../b` and `a/x/../b`, which resolve to the same path.) The previous implementation did it that way because it worked by appending each new piece of the pattern's expansion onto the prior string (which is afaik also more or less how bash works).  I suppose I could add it as an option, with the caveat that it means slightly reduced performance (because a duplicate Path object will have to be tracked multiple times).

One possible approach is add a `Path.relative(from = cwd)` method to the PathScurry path object, which would walk up from the two paths and find their nearest ancestor.  (And we could take some shortcuts there, like if it finds that its parent path's relative value is `../foo`, and its name is bar, then you know it's `../foo/bar`, so it could start out by stashing the relative paths to all the ancestors of the cwd, and the root as `/`.)

If I did that, then the default result would be that the `a/*/../b` pattern would resolve to `a/b`, without any additional work.  Then if you set `{absolute:true}` or `{realpath:true}` in the options, you'd get `/full/path/to/a/b`.

SimenB:
> If I did that, then the default result would be that the `a/*/../b` pattern would resolve to `a/b`, without any additional work. Then if you set `{absolute:true}` or `{realpath:true}` in the options, you'd get `/full/path/to/a/b`.

That seems very reasonable to me

isaacs:
For the last pathological perf case, I _think_ I can fix it with some clever preprocessing in minimatch, just need to prove that these equivalencies hold in all cases:

1. `pre/**/../<more>` is equivalent to `pre/../**/<more>`
2. `pre/<not **>/../<more>` is equivalent to `pre/<more>`

If that's the case, than I can eliminate `..` from the pattern in all cases except for the first or final position, and never have a case where it causes repeated `**` calls, because a pattern like `*/../**/*/../**` will resolve down to just `./**`

isaacs:
Ah, no, `pre/**/..` isnt equivalent to `pre/../**`, because the first will only include the parent and all children of `pre`, but the second will also include the siblings of `pre`.

isaacs:
But, I think I can still improve it a different way, by only ever walking a given path with a given pattern set one time. 

Currently, a pattern like `x/**/../*.txt` will search the `x` path for `*.txt` once for each child of `x`. (Same for their children, etc.) Should be able to short circuit that.

isaacs:
Alright! Passing with full coverage on linux and macos, passing with ~98% coverage on windows, because I'm still skipping all the symlink and realpath tests.

All that's left now is docs.

PR #506, opened 2023-03-03
chore: remove eslint-config-prettier devDep

nschonni:
Doesn't appear that ESLint is wired up right now, so the plugin shouldn't be needed


isaacs:
Just wanted to let you know I haven't forgotten about these, just been busy with actual work coding, and even though these are fairly low-impact, I want to review them carefully, because if they're worthwhile, I'll pull them into the package setup scripts I use as part of `npm init isaacs`.  I'll probably get to that in the next week or two.

isaacs:
Thanks!

PR #507, opened 2023-03-03
ci: test on Node 14

nschonni:
Although it hits EOL on 2023-04-30, I noticed it was still declared as a supported version in the `engines` so thought it made sense to still have it in the CI till it gets dropped

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,11 @@ jobs:
   build:
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 19.x]
+        node-version:
+          - 14
+          - 16
+          - 18
+          - 19
         platform:
         - os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


nschonni:
Looks like it's failing on 14 https://github.com/nschonni/node-glob/actions/runs/4321321988/jobs/7542422353
Thinks that's related to the test code not supporting 14 rather than the library

nschonni:
Worked around the first issue, but there are other failures https://github.com/nschonni/node-glob/actions/runs/4321406916/jobs/7542599293

isaacs:
Support for v14 is best effort. If you wanna do the work to get all the tests updated to work there, I'll take the patch, but it wasn't really worth the effort imo 😅

nschonni:
Fair enough. Since the EOL for that one isn't that far off, I'll just close this.

PR #511, opened 2023-03-09
fix heading in changelog file

paularmstrong:
**Problem:** There is a typo in the changelog first level heading

**Solution:** Fix it

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,4 @@
-# cganhe glo
+# changelog
 
 ## 9.2
 


isaacs:
That's not a typo

paularmstrong:
Is there some joke here? Can you give some more context? Surely I'm not the only person that sees this and is very confused.

isaacs:
Good question! I must admit I don't fully know the answer myself, or at least, not in a way I can clearly articulate, but I'll do my best.

I wouldn't call it a "joke" in the traditional sense, and the confusion is kinda the point. It's more of an artistic statement, or a koan of sorts, something surprising and inscrutable that jumps out and reminds the reader of the inherent meaninglessness and absurdity of an activity or custom. I usually jumble up the heading in my changelogs for this reason.

Writing a changelog can be helpful, but it is also a bit weird if you think about it. The changes are already either clear in the commit log, or in the case of major changes, often going to be surprising whether they're documented here or not. People tend to view it as some sort of binding contract, like it's a sin to have a breaking change in a semver major that isn't called out in a changelog, but then stop being upset if you just edit the changelog to add it. These files are either generated automatically from commit messages (and thus extraneous) or hand authored (and thus untrustworthy). The heading doesn't matter, because we all know what it is; in fact, the whole activity is absurd. But we do it anyway, because like so many social customs, it is helpful, despite the inherent weirdness.

I feel it is important and beneficial to be woken up when we are engaging in social customs that go unacknowledged and unexamined. A bizarre heading, so completely misspelled as to be beyond a simple mistake, but still somehow "wrong" and jarring, serves this purpose. And the fact that its wrongness does not in any way affect the utility/futility of the change log itself, makes it all the more so.

paularmstrong:
That's a bit of a pedantic way to say, "I think changelogs are dumb". If you don't want one, don't have one.

But also, figuring out why the order of my responses changed from v8 to v9 was a really nice callout that I could not very easily find in the git commit log. It's really difficult to sift through git commits with summaries like ["x"](https://github.com/isaacs/node-glob/commit/135a95e82140ae98b7199e278a3a2e896a50b586). So now I need to read hundreds of commits to figure out what changed in a major version.

Or I could read a helpful `changelog.md`, but I get there and the document looks like someone tried to write `change log` but was using qwerty key locations on a dvorak keyboard.

isaacs:
But I don't think they're dumb at all!  And I _do_ want one, for the very reason you found it useful, that's why I have one.  Like I said, it's hard to explain.

> Or I could read a helpful changelog.md, but I get there and the document looks like someone tried to write change log but was using qwerty key locations on a dvorak keyboard.

And in the process of doing this mundane automatic activity, you saw something surprising and strange, and it caused you to wake up and wonder.  That's what I'm saying.  That's why I do it.  Open source should be a little bit weird.  I realize my esthetic is not for everyone, but it's mine, and if I'm going to spend my precious life time creating these things and giving them away for free, it ought to be entertaining to me, don't you think?

paularmstrong:
Okay okay, I think I get it... How about `changeglob`?

isaacs:
Lol that's brilliant. Yes. https://github.com/isaacs/node-glob/commit/fa8ed5e7986ee3b29e1b437f895fbf3255f0d549

PR #513, opened 2023-03-12
Update globSync documentation to match signature

bkjohnson:
I noticed that `globSync` can also accept an array of strings:

https://github.com/isaacs/node-glob/blob/fa8ed5e7986ee3b29e1b437f895fbf3255f0d549/src/index.ts#L71-L74

This just updates the function signature shown in the Readme to match the code.


isaacs:
Thanks!

PR #528, opened 2023-05-16
(doc) Switch http to https in blog.izs.me reference

scovetta:
This PR changes http://blog.izs.me/ to https://blog.izs.me/ in the package author field. HTTPS is supported by the end point, and this shouldn't have any impact to the use of the package.


isaacs:
Wow that's old. Thanks.

PR #534, opened 2023-06-13
feat: add default cli flag

nohehf:
This PR adds the `--default=<default_pattern>` flag to glob CLI.
It allows one to pass a default pattern that will be used if no positional argument is given.
This is especially useful when glob is aliased to some command. For instance, we can have in a `package.json`:
```JSON
{
  [...],
  "scripts": {
    "test": "glob -c \"<test_command>\" --default \"./**/*.(spec|test).ts\""
    }
}
```

This allows one to pass a glob to select tests or run all tests if nothing is passed.
I currently use glob to pass files to our test command, hence the PR, but it likely has many other potential applications.
I intended to write a test, but looking at the codebase, I'm unsure if there is a way to test glob when used in CLI mode (via `src/bin.ts`). Please let me know if I'm wrong.

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -2,11 +2,11 @@
 import { foregroundChild } from 'foreground-child'
 import { existsSync } from 'fs'
 import { jack } from 'jackspeak'
-import { globStream } from './index.js'
 import { version } from '../package.json'
+import { globStream } from './index.js'
 
 const j = jack({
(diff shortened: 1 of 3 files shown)


nohehf:
Hey @isaacs any updates on this ? Thx!

nohehf:
Ty @isaacs :)

PR #546, opened 2023-08-24
fix bug in example, changed mtimeMS to mtimeMs

fALKENdk:
Updated mtimeMs casing in the timeSortedFiles example


isaacs:
Thanks!

PR #553, opened 2023-09-21
fix(types): re-export some stuff from path-scurry

### LINKED ISSUE #551: return type of glob withFileTypes is not exported
Calling `glob(x, {withFileTypes: true})` resolves with a `Path[]`, but `Path` is not exported from `glob`; that means if a consumer needs the type, e.g.:

```ts
let paths: Path[];
try {
  paths = await glob('*', {withFileTypes: true});
} catch {
  // ...
}
```

...they cannot get the `Path` type unless they either a) assume that `path-scurry` will always be resolvable since `glob` is, or b) add `path-scurry` as a direct dependency, or c):

```ts
// dude
let paths: Exclude<Awaited<ReturnType<typeof glob>>, string[]>;
```

Suggestion: `export type {Path}` somewhere in `src/index.ts`.  I can send a PR if you wish.

ty

boneskull:
This re-exports types that are accessible via the `Path` objects sometimes returned by `glob`.

Types within `path-scurry` consumed only by members tagged `@internal` are not re-exported.

* * *

Closes #551.

So, I didn't want to re-export _everything_, because a) you have to be explicit about what to re-export in TypeScript v4.9.5 using the `export type from` syntax, so I couldn't get away with it easily, and b) some portion of the exports from `path-scurry` look like they're only needed for consumers writing subclasses that extend `PathBase`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -196,6 +196,14 @@ export type {
 export { hasMagic } from './has-magic.js'
 export type { IgnoreLike } from './ignore.js'
 export type { MatchStream } from './walker.js'
+export type {
+  Path,
+  WalkOptionsWithFileTypesTrue,
+  WalkOptionsWithFileTypesUnset,
+  WalkOptions,
+  FSOption,
+} from 'path-scurry'
+
 /* c8 ignore stop */
 
 export const glob = Object.assign(glob_, {


boneskull:
(It looks like e.g., `PathBase.childrenCache` could be `protected`?)

boneskull:
CI failing on main branch

boneskull:
This also wants `FSOption`

isaacs:
Ah, yeah, some windows path issues.

PR #655, opened 2026-07-19
fix: ignore

Zamiell:
Hello again isaac!

Today, I found a type bug in node-glob. The following code results in a type error:

```ts
const MY_IGNORES = ["foo", "bar"] as const;
await glob("**/*.txt, {
  ignore: MY_IGNORES,
})
```

This PR fixes the bug. Note that this change is backwards compatible because in TypeScript, you don't have to do `readonly string[] | string[]`, because `readonly string[]` is the superset of both these things due to how the type system works.

Furthermore, the same issue seems to be present in the codebase in other places:

- All  glob* , stream, and iterate functions in  src/index.ts 
- new Glob(pattern, options)  in  src/glob.ts 
- hasMagic(pattern)  in  src/has-magic.ts

Do you want me to fix those too?


frontzhm:
我看邮件的反应很迟钝。。。如果有事发短信提醒我看邮件。。。。。。

Zamiell:
^ Seems like a spam bot. I reported it to GitHub.

